PR #4, opened 2017-11-05
Bind this to server instance

allevo:
Bind this to http server instance

The right behaviour calling the handler to the `request` event. All events are fired binding `this` to the event emitter.

### DIFF
--- a/index.js
+++ b/index.js
@@ -62,17 +62,19 @@ function inject (dispatchFunc, options, callback) {
     }
   }
 
+  const that = options.server || {}
+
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -62,17 +62,19 @@ function inject (dispatchFunc, options, callback) {
     }
   }
 
+  const that = options.server || {}
```
mcollina: can you call it `server`? `that` seems `this`.

allevo: done

PR #5, opened 2017-11-15
update request.js

makeco:
coverage `payload.length` to string

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -475,6 +475,18 @@ test('can override stream payload content-length header', (t) => {
   })
 })
 
+test('can override stream payload content-length header without request content-length', (t) => {
+  t.plan(1)
+  const dispatch = function (req, res) {
+    res.writeHead(200, { 'Content-Type': 'text/plain' })
+    res.end(req.headers['content-length'])
+  }
+
+  inject(dispatch, { method: 'post', url: '/', payload: getTestStream() }, (res) => {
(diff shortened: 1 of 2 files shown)


makeco:
ok

test/test.js
```diff
@@ -475,6 +475,18 @@ test('can override stream payload content-length header', (t) => {
   })
 })
 
+test('can override stream payload content-length header without request content-length', (t) => {
+  t.plan(1)
+  const dispatch = function (req, res) {
+    res.writeHead(200, { 'Content-Type': 'text/plain' })
+    res.end(req.headers['content-length'])
```
mcollina: You should do an assertion here rather than returning it.

makeco:
```js
const dispatch = function (req, res) {
    res.writeHead(200, { 'Content-Type': 'text/plain' })
    t.equal(req.headers['content-length'], '2')
}
```

test/test.js
```diff
@@ -475,6 +475,18 @@ test('can override stream payload content-length header', (t) => {
...
+  }
+
+  inject(dispatch, { method: 'post', url: '/', payload: getTestStream() }, (res) => {
```
mcollina: also you should check here if the `content-length` is a number.

delvedor:
Before accept this, I want to answer the question posted in https://github.com/fastify/fastify/pull/467#issuecomment-344602561.

PR #9, opened 2017-12-29
Use safe-buffer for Node 4 compatibility (fix #8)

### LINKED ISSUE #8: Buffer.from is not node 4 compatible
As titled. https://nodejs.org/api/buffer.html#buffer_class_method_buffer_from_array Fastify has support for node 4.

jsumners:
As long as Fastify intends to support Node 4 then this needs to use the `safe-buffer` module.


StarpTech:
great, LGTM

Issue #10, opened 2018-01-05
handle errors emitted through connection.destroy()


mcollina:
Unfortunately the current interface does not allow us from surfacing those errors.

I propose that we change:

1. from `inject(data, callback(res))` to `inject(data, callback(err, res))`
2. from `inject(data).then(res)` to `inject(data).catch().then()`.

delvedor:
Agree.
I'll work on this.

delvedor:
Closed in #11

PR #11, opened 2018-01-05
Added error handling

delvedor:
As titled.
This is a breaking change in the api.

Related: https://github.com/fastify/light-my-request/issues/10 and https://github.com/fastify/fastify/pull/617#issuecomment-355526614

~~@mcollina [Those](https://github.com/fastify/light-my-request/blob/handle-errors/test/test.js#L608-L650) test are failing, any idea why?~~
  
cc @fastify/fastify

### DIFF
--- a/index.js
+++ b/index.js
@@ -70,9 +70,9 @@ function inject (dispatchFunc, options, callback) {
 
     return req.prepare(() => dispatchFunc.call(server, req, res))
   } else {
-    return new Promise((resolve) => {
+    return new Promise((resolve, reject) => {
       const req = new Request(options)
-      const res = new Response(req, resolve)
+      const res = new Response(req, resolve, reject)
 
       req.prepare(() => dispatchFunc.call(server, req, res))
     })
--- a/lib/response.js
+++ b/lib/response.js
@@ -5,7 +5,7 @@ const http = require('http')
 const Stream = require('stream')
 const util = require('util')
 
-function Response (req, onEnd) {
+function Response (req, onEnd, reject) {
   http.ServerResponse.call(this, {
     method: req.method,
     httpVersionMajor: 1,
@@ -20,7 +20,27 @@ function Response (req, onEnd) {
   this.once('finish', () => {
     const res = generatePayload(this)
     res.raw.req = req
-    process.nextTick(() => onEnd(res))
+    if (typeof reject === 'function') {
+      process.nextTick(() => onEnd(res))
+    } else {
+      process.nextTick(() => onEnd(null, res))
+    }
+  })
+
+  this.connection.once('error', err => {
+    if (typeof reject === 'function') {
+      process.nextTick(() => reject(err, null))
+    } else {
+      process.nextTick(() => onEnd(err, null))
+    }
+  })
+
+  this.once('error', err => {
+    if (typeof reject === 'function') {
+      process.nextTick(() => reject(err, null))
+    } else {
+      process.nextTick(() => onEnd(err, null))
+    }
   })
 }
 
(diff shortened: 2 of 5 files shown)


lib/response.js
```diff
@@ -20,7 +20,27 @@ function Response (req, onEnd) {
   this.once('finish', () => {
     const res = generatePayload(this)
     res.raw.req = req
-    process.nextTick(() => onEnd(res))
+    if (typeof reject === 'function') {
```
jsumners: If you're doing `new Promise((resolve, reject) => {})` when would `reject` ever _not_ be a function?

delvedor: In the case you are using callback style :)
https://github.com/fastify/light-my-request/blob/aeb6c202b12262f65a692d3d80a458561e4aa7b6/index.js#L67-L71

jsumners: Oh, I missed that first `if` case. 👍

mcollina:
They are failing because you should reply. Or call `destroy()` on the res connection.

delvedor:
@mcollina but why before this pr that tests were passing?
Here I've only added an event listener.

delvedor:
@mcollina you're right. I've updated the test. Thanks!

lib/response.js
```diff
@@ -20,7 +20,27 @@ function Response (req, onEnd) {
...
+      process.nextTick(() => onEnd(res))
+    } else {
+      process.nextTick(() => onEnd(null, res))
+    }
+  })
+
+  this.connection.once('error', err => {
```
StarpTech: Connection issues should be covered by a test too.

lib/response.js
```diff
@@ -20,7 +20,27 @@ function Response (req, onEnd) {
...
+    if (typeof reject === 'function') {
+      process.nextTick(() => reject(err, null))
```
StarpTech: Need to pass `null` ?

lib/response.js
delvedor: Connection errors are covered by
https://github.com/fastify/light-my-request/blob/ad9cc6e87df1235e92ffbd8ec38853245195386c/test/test.js#L757-L777

The res error has not test at the moment.

index.js
```diff
@@ -70,9 +70,9 @@ function inject (dispatchFunc, options, callback) {
 
     return req.prepare(() => dispatchFunc.call(server, req, res))
   } else {
-    return new Promise((resolve) => {
+    return new Promise((resolve, reject) => {
       const req = new Request(options)
-      const res = new Response(req, resolve)
+      const res = new Response(req, resolve, reject)
```
nwoltman: I'd like to suggest doing something like this:

```js
const res = new Response(req, (err, res) => {
  if (err) reject(err)
  else resolve(res)
})
```

That way you wouldn't need the `if (this._promiseCallback)` checks in `lib/response.js` so `onEnd` could always just be a regular callback.

Up to you though :)

delvedor: Pr are welcome :)

StarpTech:
Why do you pass `null` to any reject function?

delvedor:
@StarpTech where?

StarpTech:
@delvedor nevermind this was already fixed .. browser cache or so :D

lib/response.js
```diff
@@ -2,10 +2,10 @@
 
 const Buffer = require('safe-buffer').Buffer
 const http = require('http')
-const Stream = require('stream')
+const stream = require('readable-stream')
```
allevo: Why this change?

delvedor: Compatibility down to node v4.
Socket.destroy works natively only from node v8.

allevo: As always, thanks!

PR #18, opened 2018-03-25
Remove `readable-stream`.

wtgtybhertgeghgtwtg:
`lib/request.js` uses `stream`, while `lib/response.js` uses `readable-stream`.  Unless I'm missing something, it should probably be one way or the other.

### DIFF
--- a/lib/response.js
+++ b/lib/response.js
@@ -2,7 +2,7 @@
 
 const Buffer = require('safe-buffer').Buffer
 const http = require('http')
-const stream = require('readable-stream')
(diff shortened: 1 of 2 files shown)


wtgtybhertgeghgtwtg:
Guess that fails in older `node` versions.  Sorry for the bother.

mcollina:
Ideally they should both use readable-stream.

wtgtybhertgeghgtwtg:
I'd file a PR for that, but I'm wary of `instanceof` checks applied to user input from a library class.

PR #20, opened 2018-04-21
Only http.METHODS

allevo:
From https://github.com/fastify/fastify/pull/883#issuecomment-382182010

### DIFF
--- a/index.js
+++ b/index.js
@@ -42,7 +42,7 @@ const schema = {
     },
     authority: { type: 'string' },
     remoteAddress: { type: 'string' },
-    method: { type: 'string', enum: http.METHODS },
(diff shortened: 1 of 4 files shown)


allevo:
Truly the method http is case-sensitive by RFC

https://tools.ietf.org/html/rfc2616#section-5.1.1

This is a bugfix. Anyway if you want, I change in order to allow both kind of cases.

delvedor:
I know, in fact as you can see we uppercase it internally.
Unfortunately this bugfix is also a breaking change, so it is better to keep both uppercase and lowercase.

allevo:
I have added a test for the lower case HTTP method. So that behaviour is tracked

allevo:
I've no rights for merging this

PR #24, opened 2018-07-06
Force stream resume

allevo:
As titled, if the payload is a stream, `light-my-request` should force its resume.

Fix https://github.com/fastify/fastify/pull/1026

### DIFF
--- a/lib/request.js
+++ b/lib/request.js
@@ -84,7 +84,7 @@ Request.prototype.prepare = function (next) {
     return next()
   })
 
-  // Force to resume the stream. Need for stream1
+  // Force to resume the stream. Needed for Stream 1
   this._lightMyRequest.payload.resume()
 }
 
--- a/lib/request.js
+++ b/lib/request.js
@@ -84,6 +84,7 @@ Request.prototype.prepare = function (next) {
     return next()
   })
 
+  // Force to resume the stream. Need for stream1
   this._lightMyRequest.payload.resume()
 }
 
--- a/lib/request.js
+++ b/lib/request.js
@@ -83,6 +83,8 @@ Request.prototype.prepare = function (next) {
     this._lightMyRequest.payload = payload
     return next()
   })
+
+  this._lightMyRequest.payload.resume()
 }
 
 Request.prototype._read = function (size) {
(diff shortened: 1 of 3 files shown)


test/test.js
```diff
@@ -797,6 +797,35 @@ test('HTTP method is case insensitive', (t) => {
   })
 })
 
+const FormData = require('form-data')
```
mcollina: Can you move this to the top of the file, and add form-data to devDependencies?

lib/request.js
```diff
@@ -83,6 +83,8 @@ Request.prototype.prepare = function (next) {
     this._lightMyRequest.payload = payload
     return next()
   })
+
+  this._lightMyRequest.payload.resume()
```
mcollina: Can you add a comment before this line? This is happening because the payload is a stream1

allevo: `form-data` inherits indirectly from `require('stream').Stream`

https://github.com/form-data/form-data/blob/master/lib/form_data.js#L16

https://github.com/felixge/node-combined-stream/blob/master/lib/combined_stream.js#L18

is `require('stream').Stream` the stream1?

test/test.js
allevo: ops... sorry! done

lib/request.js
mcollina: yes it is.

test/test.js
```diff
@@ -797,6 +799,34 @@ test('HTTP method is case insensitive', (t) => {
...
+    req.on('data', d => {
+      body += d
+    })
+    req.on('end', () => {
+      res.end(body)
+    })
+  }
+
+  const form = new FormData()
+  form._boundary = 'the-boundary'
```
mcollina: I would not rely on this `_boundary` property, it makes the test less reliable.

allevo: done

lib/request.js
```diff
@@ -83,6 +83,9 @@ Request.prototype.prepare = function (next) {
     this._lightMyRequest.payload = payload
     return next()
   })
+
+  // Force to resume the stream. Need for stream1
```
mcollina: Need -> needed
stream1 -> Streams 1

allevo: done

PR #32, opened 2019-01-22
Add `query` option

arniu:
See [Help #46](https://github.com/fastify/help/issues/46)

### DIFF
--- a/lib/request.js
+++ b/lib/request.js
@@ -1,20 +1,48 @@
 'use strict'
 
 const { Readable } = require('readable-stream')
-const Url = require('url')
+const qs = require('querystring')
 const util = require('util')
+const url = require('url')
 
-function Request (options) {
-  Readable.call(this)
-  // options: method, url, payload, headers, remoteAddress
-  var url = options.url
-  if (typeof url === 'object') {
-    url = Url.format(url)
+function parseURL (options) {
+  const parsedURL = url.parse(
+    typeof options.url === 'object'
+      ? url.format(options.url)
+      : options.url
+  )
+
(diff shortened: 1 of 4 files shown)


lib/request.js
```diff
@@ -1,19 +1,51 @@
 'use strict'
 
 const { Readable } = require('readable-stream')
-const Url = require('url')
+const qs = require('querystring')
 const util = require('util')
+const url = require('url')
+
+function parseURL (options) {
+  const parsedURL = url.parse(
```
delvedor: As far as I know, `url.parse` is deprecated, you should use WHATWG URL API instead :)

mcollina: You should use `const { URL} = require('url')` and then `new URL()`.

arniu: The new WHATWG `URL` changes a lot. 

For example, `'/example?q=1'` is an invalid URL now:

```
new URL('/example?q=1')
{ TypeError [ERR_INVALID_URL]: Invalid URL: /example?q=1
    at onParseError (internal/url.js:240:17)
    at parse (internal/url.js:249:3)
    at new URL (internal/url.js:324:5) input: '/example?q=1' }
```

As a result, a `base` option may be required.

So, I thinks it's another task to mitigate to the new WHATWG `URL`.

mcollina: I agree

arniu:
@delvedor Would you please merge it?

delvedor:
Hello! Sorry for the delay.
Do you want to work right away for the URL parsing migration? In that case, I'll wait before release a new version :)

PR #33, opened 2019-01-28
Migrate to WHATWG URL

arniu:


lib/parseURL.js
```diff
@@ -0,0 +1,47 @@
...
+ * @param {Object} options
+ * @param {(Object|String)} options.url
+ * @param {Object} [options.query]
+ * @return {URL}
+ */
+module.exports = function parseURL (options) {
+  let parsedURL = null
+  let mergedQuery = {}
+
+  if (typeof options.url === 'object') {
```
mcollina: Is this case possible? From the README, we are not supporting this.

arniu: According the schema, `url` may be `String` or `Object`.

lib/request.js
```diff
@@ -1,29 +1,20 @@
...
+const parseURL = require('./parseURL')
+
+/**
+ * Get hostname:port
+ *
+ * @param {URL} parsedURL
+ * @return {String}
+ */
+function hostHeaderFromURL (parsedURL) {
+  return parsedURL.port
```
delvedor: What are you checking here?
The existence of the `port` value?
If so, I would add a stricter check.

arniu: The new URL's `port` will be empty when using default ports:

```javascript
new URL('http://localhost').port === ''
new URL('http://localhost:80').port === ''
new URL('https://localhost').port === ''
new URL('https://localhost:443').port === ''
```

PR #39, opened 2019-03-21
Add TypeScript typings

Ethan-Arrowood:
Spent some time trying to complete this tonight. Please review as accurately as possible. All of my types are from reading the code. I based some off of the JSON Schema object but also included anything written in the documentation too. There were a couple `any` types I'd like to see if we can make more specific, as well as some general questions I have: 

- Is the `port` property on the `InjectOptions.url` type of `string | number`?
- What is `query`? Could this be typed as `string`?
- What types can `payload` be? I've implemented it as a generic for now but I think we can scope this to an union type list: `string | object | Buffer | NodeJS.ReadableStream` (this is actually copied from the current `fastify.d.ts` types `HTTPInjectOptions` so I believe it should be safe to implement here as well). Depending on what happens here I'll most likely be removing the generic.
- Am I correct that `Response._lightMyRequest.payloadChunks` is of type `Array<Buffer>`?

Not ready to be merged yet as I'm still playing with `tsd` to test the new types. It is a young module with a tiny API so it might not be quite ready yet, but I'd like to revisit it with a clear head tomorrow.

Finally, I need to experiment with the declaration file exports, namespace, and some of the internal TSC stuff I used (such as `interface`) to make sure it is correct.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,15 +1,28 @@
 import * as stream from 'stream'
 import * as http from 'http'
 
-type HTTPMethods = 'DELETE' | 'GET' | 'HEAD' | 'PATCH' | 'POST' | 'PUT' | 'OPTIONS'
+type HTTPMethods = 'DELETE' | 'delete' |
+                   'GET' | 'get' |
+                   'HEAD' | 'head' |
+                   'PATCH' | 'patch' |
+                   'POST' | 'post' |
+                   'PUT' | 'put' |
+                   'OPTIONS' | 'options'
 
 declare namespace LightMyRequest {
-
-  type inject<Payload> = (
-    dispatchFunc: (req: Request, res: Response) => void,
+  function inject (
+    dispatchFunc: DispatchFunc,
+    options: string | InjectOptions
+  ): Promise<Response>
+  function inject (
+    dispatchFunc: DispatchFunc,
     options: string | InjectOptions,
-    callback: (err: Error, payload: any) => void | Promise<Payload>
-  ) => void
+    callback: (err: Error, response: Response) => void
+  ): void
+
+  type DispatchFunc = (req: Request, res: Response) => void
+
+  type InjectPayload = string | object | Buffer | NodeJS.ReadableStream
 
   function isInjection (obj: Request | Response): boolean
 
@@ -18,11 +31,11 @@ declare namespace LightMyRequest {
       pathname: string
       protocal?: string
       hostname?: string
-      port?: any
-      query?: any
+      port?: string | number
+      query?: string
     }
     headers?: http.IncomingHttpHeaders | http.OutgoingHttpHeaders
-    query?: any
+    query?: string
     simulate?: {
       end: boolean,
       split: boolean,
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -7,17 +7,21 @@
...
+    "tap": "^12.5.1",
+    "tsd": "^0.7.0"
   },
   "scripts": {
     "test": "npm run lint && npm run unit",
     "lint": "standard",
     "unit": "tap test/test.js",
-    "coverage": "npm run unit -- --cov --coverage-report=html"
+    "coverage": "npm run unit -- --cov --coverage-report=html",
+    "test:tsd": "tsd"
```
mcollina: How does this work?

index.d.ts
```diff
@@ -0,0 +1,72 @@
...
+
+  function isInjection (obj: Request | Response): boolean
+
+  interface InjectOptions {
+    url: string | {
+      pathname: string
+      protocal?: string
+      hostname?: string
+      port?: any
+      query?: any
```
mcollina: I think this is string.

index.d.ts
```diff
@@ -0,0 +1,72 @@
...
+  ) => void
+
+  function isInjection (obj: Request | Response): boolean
+
+  interface InjectOptions {
+    url: string | {
+      pathname: string
+      protocal?: string
+      hostname?: string
+      port?: any
```
mcollina: number or string

index.d.ts
```diff
@@ -0,0 +1,72 @@
...
+  type inject<Payload> = (
+    dispatchFunc: (req: Request, res: Response) => void,
+    options: string | InjectOptions,
+    callback: (err: Error, payload: any) => void | Promise<Payload>
+  ) => void
+
+  function isInjection (obj: Request | Response): boolean
+
+  interface InjectOptions {
+    url: string | {
```
mcollina: I’m not sure we can really pass an object here. Can you check in code?

index.d.ts
```diff
@@ -0,0 +1,72 @@
...
+    server?: http.Server
+  }
+
+  interface Request<Payload = string> extends stream.Readable {
+    url: string
+    httpVersion: string
+    method: HTTPMethods
+    headers: http.IncomingHttpHeaders
+    connection: string
+    _lightMyRequest: {
```
mcollina: This should not be exposed.

index.d.ts
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  interface Response extends http.ServerResponse {
+    _lightMyRequest: {
+      headers: http.OutgoingHttpHeaders,
+      trailers: { [key: string]: string },
+      payloadChunks: Array<Buffer>
+    }
+    _headers: http.OutgoingHttpHeaders
+    _promiseCallback: boolean
```
mcollina: These should not be exposed

index.d.ts
Ethan-Arrowood: Based on the schema in `index.js`:
https://github.com/fastify/light-my-request/blob/f06c85f5c3f3dc98ac9adc37d7d0c2ad1cc5824a/index.js#L16-L27

package.json
Ethan-Arrowood: Sorry this is incomplete. `tsd` is the typescript definition tester. All it needs is this command and then the `tsd: {}` options further down the file. It runs any file with `.test-d.ts` located in the `test` directory. I'm still working with it so I'll push the test file later

index.d.ts
Ethan-Arrowood: ping @mcollina for updates. Pinging just in case this notification slipped through; I know you can get very busy so no rush on this! Thank you 😁

tsconfig.json
```diff
@@ -0,0 +1,13 @@
+{
+  "compilerOptions": {
+    "target": "es6",
+    "module": "commonjs",
+    "noEmit": true,
+    "strict": true,
+    "noImplicitAny": true
+  },
+  "files": [
+    "./test.ts",
```
Ethan-Arrowood: Need to remove this

Ethan-Arrowood:
Pinging @delvedor for review in case the notification was missed

Issue #50, opened 2019-09-12
Provide RESTful and chainable apis to make things easier


fralonra:
## 🚀 Feature Proposal

Give direct apis for every HTTP methods and others like `header` and `body`.

## Motivation

Currently we use properties in `options` object passed to `inject` to tell light-my-request what is the request url, what is the method, and some other things.

I think it would make coding simpler if we provide standalone apis for such things, and make them chainable. As a result, the `options` can be omitted, and the callback can be moved to the last method of the chain.

## Example

```js
const inject = require('light-my-request')

const dispatch = (req, res) => {
  // some code
}

inject(dispatch)
  .get('/test')
  .header({ foo: 'bar' })
  .end((err, res) => {
    // some debug code
  })

inject(dispatch)
  .post('/test')
  .body({ foo: 'bar' })
  .end((err, res) => {
    // some debug code
  })

// you can follow the old way as well
inject(dispatch, { method: 'get', url: '/test' }, (err, res) => {
    // some debug code
  })
```

fralonra:
@mcollina 
Did you mean parse JSON in the request body? It's already handled by `body` and `payload`.

fralonra:
To avoid messing up the original code, I decided to make another library that wraps `light-my-request` as a first step.

Here it is. [injectar](https://github.com/fralonra/injectar).

mcollina:
Awesome work! I’m closing this then.

PR #53, opened 2019-10-08
Chainable api support

fralonra:
The original issue is https://github.com/fastify/light-my-request/issues/50.

And I'm glad to update the document and typings if this feature is accepted.

#### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,21 @@ try {
 }
 ```
 
+You can also use chaining methods if you do not pass the callback function. Check [here](#method-chaining) for details.
+
+```js
+// chaining methods
+inject(dispatch)
+  .get('/')                   // set the request method to GET, and request URL to '/'
+  .headers({ foo: 'bar' })    // set the request headers
+  .payload('request payload') // set the request payload
+  .body('request body')       // alias for payload
+  .query({ foo: 'bar' })      // set the query parameters
+  .end((err, res) => {
+    console.log(res.payload)
+  })
+```
+
 File uploads (multipart/form-data) can be achieved by using [form-data](https://github.com/form-data/form-data) package as shown below:
 
 ```js
@@ -103,7 +118,7 @@ The declaration file exports types for the following parts of the API:
 
 ## API
 
-#### `inject(dispatchFunc, options, callback)`
+#### `inject(dispatchFunc[, options, callback])`
 
 Injects a fake request into an HTTP server.
 
@@ -146,6 +161,15 @@ Note: You can also pass a string in place of the `options` object as a shorthand
 #### `inject.isInjection(obj)`
 
 Checks if given object `obj` is a *light-my-request* `Request` object.
+
+#### Method chaining
+
+There are following methods you can used as chaining:
+- `delete`, `get`, `head`, `options`, `patch`, `post`, `put`, `trace`. They will set the HTTP request method and also the request URL.
+- `body`, `headers`, `payload`, `query`. They can be used to set the request options object.
+And finally you need to call `end`. It has the signature `function (callback)`.
+
+Note: The application would not response multiple times. If you try to do so by calling `end()` twice for example, the application would throw an error.
 
 ## Acknowledgements
 This project has been forked from [`hapi/shot`](https://github.com/hapijs/shot) because we wanted to support *Node ≥ v4* and not only *Node ≥ v8*.  
--- a/README.md
+++ b/README.md
@@ -168,7 +168,42 @@ Checks if given object `obj` is a *light-my-request* `Request` object.
 There are following methods you can used as chaining:
 - `delete`, `get`, `head`, `options`, `patch`, `post`, `put`, `trace`. They will set the HTTP request method and also the request URL.
 - `body`, `headers`, `payload`, `query`. They can be used to set the request options object.
+
 And finally you need to call `end`. It has the signature `function (callback)`.
+If you invoke `end` without a callback function, the method will return a promise, thus you can:
+
+```js
+const chain = inject(dispatch).get('/')
+
+try {
+  const res = await chain.end()
+  console.log(res.payload)
+} catch (err) {
+  // handle error
+}
+
+// or
+chain.end()
+  .then(res => {
+    console.log(res.payload)
+  })
+  .catch(err => {
+    // handle error
+  })
+```
+
+By the way, you can also use promises without calling `end`!
+
+```js
+inject(dispatch)
+  .get('/')
+  .then(res => {
+    console.log(res.payload)
+  })
+  .catch(err => {
+    // handle error
+  })
+```
 
 Note: The application would not response multiple times. If you try to do so by calling `end()` twice for example, the application would throw an error.
 
--- a/README.md
+++ b/README.md
@@ -50,9 +50,15 @@ You can also use chaining methods if you do not pass the callback function. Chec
 inject(dispatch)
   .get('/')                   // set the request method to GET, and request URL to '/'
   .headers({ foo: 'bar' })    // set the request headers
+  .query({ foo: 'bar' })      // set the query parameters
+  .end((err, res) => {
+    console.log(res.payload)
+  })
+
+inject(dispatch)
+  .post('/')                  // set the request method to POST, and request URL to '/'
   .payload('request payload') // set the request payload
   .body('request body')       // alias for payload
-  .query({ foo: 'bar' })      // set the query parameters
   .end((err, res) => {
     console.log(res.payload)
   })
--- a/README.md
+++ b/README.md
@@ -62,6 +62,15 @@ inject(dispatch)
   .end((err, res) => {
     console.log(res.payload)
   })
+
+// async-await is also supported
+try {
+  const chain = inject(dispatch).get('/')
+  const res = await chain.end()
+  console.log(res.payload)
+} catch (err) {
+  console.log(err)
+}
 ```
 
 File uploads (multipart/form-data) can be achieved by using [form-data](https://github.com/form-data/form-data) package as shown below:
--- a/README.md
+++ b/README.md
@@ -205,7 +205,7 @@ inject(dispatch)
   })
 ```
 
-Note: The application would not response multiple times. If you try to do so by calling `end()` twice for example, the application would throw an error.
+Note: The application would not respond multiple times. If you try to invoking any method after the application has responded, the application would throw an error.
 
 ## Acknowledgements
 This project has been forked from [`hapi/shot`](https://github.com/hapijs/shot) because we wanted to support *Node ≥ v4* and not only *Node ≥ v8*.  
--- a/index.js
+++ b/index.js
@@ -63,6 +63,14 @@ const schema = {
 const optsValidator = ajv.compile(schema)
 
 function inject (dispatchFunc, options, callback) {
+  if (typeof callback === 'undefined') {
+    return new Chain(dispatchFunc, options)
+  } else {
+    return doInject(dispatchFunc, options, callback)
+  }
+}
+
+function doInject (dispatchFunc, options, callback) {
   options = (typeof options === 'string' ? { url: options } : options)
 
   if (options.validate !== false) {
@@ -89,6 +97,57 @@ function inject (dispatchFunc, options, callback) {
     })
   }
 }
+
+function Chain (dispatch, option = {}) {
+  this.option = {
+    ...option
+  }
+  this.dispatch = dispatch
+}
+
+const httpMethods = [
+  'delete',
+  'get',
+  'head',
+  'options',
+  'patch',
+  'post',
+  'put',
+  'trace'
+]
+
+httpMethods.forEach(method => {
+  Chain.prototype[method] = function (url) {
+    this.option.url = url
+    this.option.method = method.toUpperCase()
+    return this
+  }
+})
+
+const chainMethods = [
+  'body',
+  'headers',
+  'payload',
+  'query'
+]
+
+chainMethods.forEach(method => {
+  Chain.prototype[method] = function (value) {
+    this.option[method] = value
+    return this
+  }
+})
+
+Chain.prototype.end = function (callback) {
+  doInject(this.dispatch, this.option, callback)
+}
+
+Object.getOwnPropertyNames(Promise.prototype).forEach(method => {
+  if (method === 'constructor') return
+  Chain.prototype[method] = function (...args) {
+    return doInject(this.dispatch, this.option)[method](...args)
+  }
+})
 
 function isInjection (obj) {
   return (obj instanceof Request || obj instanceof Response)
--- a/index.js
+++ b/index.js
@@ -118,6 +118,9 @@ const httpMethods = [
 
 httpMethods.forEach(method => {
   Chain.prototype[method] = function (url) {
+    if (this._hasInvoked === true || this._promise) {
+      throwIfAlreadyInvoked()
+    }
     this.option.url = url
     this.option.method = method.toUpperCase()
     return this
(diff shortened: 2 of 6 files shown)


index.js
```diff
@@ -90,6 +98,57 @@ function inject (dispatchFunc, options, callback) {
...
+  doInject(this.dispatch, this.option, callback)
+}
+
+Object.getOwnPropertyNames(Promise.prototype).forEach(method => {
+  if (method === 'constructor') return
+  Chain.prototype[method] = function (...args) {
+    return doInject(this.dispatch, this.option)[method](...args)
+  }
+})
+
```
fralonra: Provides backward compatibility for promise.
Tell me if there is a better way!

index.js
```diff
@@ -90,6 +98,57 @@ function inject (dispatchFunc, options, callback) {
...
+Chain.prototype.end = function (callback) {
+  doInject(this.dispatch, this.option, callback)
+}
+
+Object.getOwnPropertyNames(Promise.prototype).forEach(method => {
+  if (method === 'constructor') return
+  Chain.prototype[method] = function (...args) {
+    return doInject(this.dispatch, this.option)[method](...args)
+  }
+})
```
fralonra: Provides backward compatibility for promise.
Tell me if there is a better way!

index.js
fralonra: Provides backward compatibility for promise.
Tell me if there is a better way!

index.js
fralonra: Provides backward compatibility for promise.
Tell me if there is a better way!

mcollina: I think this is ok. I would just have added a `then` method, but this is better.

The implementation has a side effect, as invoking multiple times `.then()` will issue multiple calls vs waiting for the same one. You'll have to get hold of a promise and the keep calling the methods on that one instead, i.e. the same `Chain` object wraps only one promise.

test/test.js
```diff
@@ -840,7 +851,17 @@ test('should throw on unknown HTTP method', (t) => {
   t.plan(1)
   const dispatch = function (req, res) { }
 
-  t.throws(() => inject(dispatch, { method: 'UNKNOWN_METHOD', url: 'http://example.com:8080/hello' }), Error)
+  t.throws(() => inject(dispatch, { method: 'UNKNOWN_METHOD', url: 'http://example.com:8080/hello' }, (err, res) => {
+    t.ok(err)
+  }), Error)
+})
```
mcollina: Why this change?

fralonra: Because with the new api, if there is no callback function, it would be treated as using chainable apis (as shown in the next test).

fralonra: I believe that it would be better to add `.json()` handler of `res` in another PR, for it has no concern with chainable apis. Do you agree?

index.js
fralonra: Got it. But I'm wondered is there any cases should we write codes like that?

However, if different `then` returns different Promise, we can test different cases using one `Chain` Instance, just like:
```js
const chain = inject(dispatch)
chain.get('/').then()
chain.post('/').then()
```
I think the above case might be more common.

mcollina: I think we should throw in the above case, you can’t set the method twice

mcollina:
@fralonra any update on this?

mcollina:
Would you mind adding a test for the multiple promise fix? Also, this would need a rebase.

fralonra:
I tried to write test like below:
```js
test('chainable api: multiple call of then should returns the same promise', (t) => {
  t.plan(1)

  function dispatch (req, res) {
    res.writeHead(200, { 'Content-Type': 'text/plain' })
    res.end('hello')
  }

  const chain = inject(dispatch).get('/')
  chain.then(res => {
    chain.then(rep => {
      t.equal(res, rep)
    })
  })
})
```
But it finally outpus the error `not ok test unfinished`.

fralonra:
There is another consideration. Currently the usage of `end()` may conflict with `then()`.

`end()` will invoke the injection as many times as this method being called, so that we can use only one `Chain` to test different cases. That's exactly what I wanted at the begining:
```js
function dispatch (req, res) {
  // has different logic for route '/one-route' and '/another-route'
}

const chain = inject(dispatch)
chain.get('/one-route').end(() => {
  // test code
})
chain.get('/another-route').end(() => {
  // test code
})
```

Obviously it conflicts with the design that multiple `then()` returns the same promise.

mcollina:
I think we should throw if end() is called multiple times.

index.js
```diff
@@ -90,6 +98,59 @@ function inject (dispatchFunc, options, callback) {
...
+  doInject(this.dispatch, this.option, callback)
+}
+
+Object.getOwnPropertyNames(Promise.prototype).forEach(method => {
+  if (method === 'constructor') return
+  Chain.prototype[method] = function (...args) {
+    if (!this._promise) {
+      this._promise = doInject(this.dispatch, this.option)[method](...args)
+    }
+    return this._promise
```
mcollina: you need to call the method on promise, passing the args to it anyway.

fralonra: Sorry for forgetting changing the code.

fralonra:
> I think we should throw if end() is called multiple times.

Should we just throw the error, or use the callback to wrap it?

mcollina:
I think we should just throw.

index.js
```diff
@@ -138,16 +139,21 @@ chainMethods.forEach(method => {
 })
 
 Chain.prototype.end = function (callback) {
+  if (this._hasInvoked === true) {
+    throw new Error('The dispatch function has already been invoked')
+  }
   doInject(this.dispatch, this.option, callback)
+  this._hasInvoked = true
```
mcollina: Can you make the chain return itself here? So that we can:

```js
await chain.end()
```

index.js
```diff
@@ -138,16 +139,21 @@ chainMethods.forEach(method => {
 })
 
 Chain.prototype.end = function (callback) {
+  if (this._hasInvoked === true) {
+    throw new Error('The dispatch function has already been invoked')
+  }
   doInject(this.dispatch, this.option, callback)
```
mcollina: we should probably store the returned promise here.

index.js
```diff
@@ -139,17 +139,25 @@ chainMethods.forEach(method => {
...
+  return this
 }
 
 Object.getOwnPropertyNames(Promise.prototype).forEach(method => {
   if (method === 'constructor') return
   Chain.prototype[method] = function (...args) {
     if (!this._promise) {
+      if (this._hasInvoked === true) {
+        throw new Error('The dispatch function has already been invoked')
+      }
```
mcollina: This is not correct. The difference between a promise and end is that one would throw, and one does not, because promises keep the resolved value.

fralonra: Since `end()` would throw if called multiple times, and multiple `then()` returns only one promise, what should we do if one calls `then()` after invoking `end()`? Should we just reject here?

Or these two methods do not effect each other, one can call `then()` after `end()` as normal, v.v.

mcollina: If inside end() you store the promise inject() returns, you can return it here.

fralonra: That's what exactly these method do currently. If you pass no callback function to `end`, it would store the promise, and return the promise in the next `then` call.

fralonra: That's what exactly these method do currently. If you pass no callback function to end, it would store the promise, and return the promise in the next then call.

Is there anything I misunderstand?

mcollina: .end() needs to set the promise as well.

fralonra: Even if there is already a callback?

mcollina: That’s an interesting case, as mixing promises and callbacks would not go well in application. I think we should just set the promise if end() is called without a callback. I would also make the error more explicit to cover the above case.

fralonra: The above case has already been handled. The application would store the promise if `end()` is called without a callback, and then we can access the promise through `then()`.

Eomm:
@fralonra Could you check the conflicts plz?

README.md
```diff
@@ -147,6 +162,15 @@ Note: You can also pass a string in place of the `options` object as a shorthand
 
 Checks if given object `obj` is a *light-my-request* `Request` object.
 
+#### Method chaining
+
+There are following methods you can used as chaining:
+- `delete`, `get`, `head`, `options`, `patch`, `post`, `put`, `trace`. They will set the HTTP request method and also the request URL.
+- `body`, `headers`, `payload`, `query`. They can be used to set the request options object.
+And finally you need to call `end`. It has the signature `function (callback)`.
```
mcollina: end() can be invoked without a callback

fralonra: That's right. But I think that's not the way we should recommend users to do, so I would not mention it in the document.

If invoking `end()` without a callback, users should access the promise through methods like `then()` later on:
```js
inject(dispatch)
  .get('/')
  .end()
  .then(res => {})
```

It acts the same as:
```js
inject(dispatch)
  .get('/')
  .then(res => {})
```
So I think there is no need to call `end()`. We just provide a way that makes the application more flexible.

index.js
delvedor: I agree, `end` should not return `this`, because you shouldn't be able to set more things once you send the request.
At the moment, this is valid:
```js
inject(dispatch)
  .get('/')                   // set the request method to GET, and request URL to '/'
  .headers({ foo: 'bar' })    // set the request headers
  .payload('request payload') // set the request payload
  .body('request body')       // alias for payload
  .query({ foo: 'bar' })      // set the query parameters
  .end((err, res) => {
    console.log(res.payload)
  })
  .query({ foo: 'bar' }
```

fralonra: If `end()` don't return `this`, then I think we should always set the callback function in `end()`, thus there is no need to store the promise in `end()`, which is requested in https://github.com/fastify/light-my-request/pull/53#discussion_r335347129

Eomm: Why we don't return `this._promise`?

In this way:
- we stop the chaining
- we can `const res = await chain.end()`

fralonra: @Eomm Great idea! I agree.

index.js
```diff
@@ -90,6 +98,73 @@ function inject (dispatchFunc, options, callback) {
...
+  'head',
+  'options',
+  'patch',
+  'post',
+  'put',
+  'trace'
+]
+
+httpMethods.forEach(method => {
+  Chain.prototype[method] = function (url) {
```
mcollina: I think we should throw if this is called after the function is invoked.

index.js
```diff
@@ -90,6 +98,73 @@ function inject (dispatchFunc, options, callback) {
...
+const chainMethods = [
+  'body',
+  'headers',
+  'payload',
+  'query'
+]
+
+chainMethods.forEach(method => {
+  Chain.prototype[method] = function (value) {
+    this.option[method] = value
```
mcollina: Also this should throw

README.md
```diff
@@ -43,6 +43,21 @@ try {
 }
 ```
 
+You can also use chaining methods if you do not pass the callback function. Check [here](#method-chaining) for details.
+
+```js
+// chaining methods
+inject(dispatch)
+  .get('/')                   // set the request method to GET, and request URL to '/'
```
delvedor: I would write two examples, one with `get` and one with `post`, since this one is not correct (you can't have a body in a `GET` request).

test/index.test-d.ts
```diff
@@ -21,6 +21,15 @@ inject(dispatch, { method: 'get', url: '/' }, (err, res) => {
   expectType<Function>(res.json)
 })
 
-expectType<Promise<Response>>(inject(dispatch, { method: 'get', url: '/' }))
+inject(dispatch)
+  .get('/')
+  .end((err, res) => {
+    expectType<Error>(err)
```
delvedor: Here you can use `expectError` :)

test/test.js
```diff
@@ -909,6 +930,271 @@ test('Should throw if both path and url are missing', (t) => {
...
+  inject(dispatch)
+    .get('/')
+    .catch(err => t.ok(err))
+})
+
+test('chainable api: multiple call of then should returns the same promise', (t) => {
+  t.plan(1)
+
+  function dispatch (req, res) {
+    res.writeHead(200, { 'Content-Type': 'text/plain' })
```
delvedor: Can you also add a check to be 100% sure that dispatch has been called only once?

test/index.test-d.ts
fralonra: I just followed L18 to keep the coding style consistent. If using `expectError` is better, I suggest to replace these all.

test/test.js
```diff
@@ -956,18 +956,19 @@ test('chainable api: backwards compatibility for promise (catch)', (t) => {
...
 })
 
-test('chainable api: multiple call of then should returns the same promise', (t) => {
+test('chainable api: multiple call of then should return the same promise', (t) => {
   t.plan(1)
+  let id = 0
 
   function dispatch (req, res) {
-    res.writeHead(200, { 'Content-Type': 'text/plain' })
+    res.writeHead(200, { 'Content-Type': 'text/plain', 'Request-Id': id++ })
```
mcollina: can you add a `t.pass()` here and increase the assertion count as well?

README.md
```diff
@@ -43,6 +43,27 @@ try {
...
+    console.log(res.payload)
+  })
+
+inject(dispatch)
+  .post('/')                  // set the request method to POST, and request URL to '/'
+  .payload('request payload') // set the request payload
+  .body('request body')       // alias for payload
+  .end((err, res) => {
+    console.log(res.payload)
+  })
```
mcollina: It would be good to add an example with await here

index.js
```diff
@@ -90,11 +98,88 @@ function inject (dispatchFunc, options, callback) {
...
+  'patch',
+  'post',
+  'put',
+  'trace'
+]
+
+httpMethods.forEach(method => {
+  Chain.prototype[method] = function (url) {
+    if (this._hasInvoked === true || this._promise) {
+      throwIfAlreadyInvoked()
```
mcollina: This is not ok, you are missing a return statement.

index.js
```diff
@@ -90,11 +98,88 @@ function inject (dispatchFunc, options, callback) {
...
+  'body',
+  'headers',
+  'payload',
+  'query'
+]
+
+chainMethods.forEach(method => {
+  Chain.prototype[method] = function (value) {
+    if (this._hasInvoked === true || this._promise) {
+      throwIfAlreadyInvoked()
```
mcollina: missing return

index.js
```diff
@@ -90,11 +98,88 @@ function inject (dispatchFunc, options, callback) {
...
+    }
+    this.option[method] = value
+    return this
+  }
+})
+
+Chain.prototype.end = function (callback) {
+  if (this._hasInvoked === true || this._promise) {
+    throwIfAlreadyInvoked()
```
mcollina: missing return

index.js
```diff
@@ -90,11 +98,88 @@ function inject (dispatchFunc, options, callback) {
...
+    return this._promise
+  }
+}
+
+Object.getOwnPropertyNames(Promise.prototype).forEach(method => {
+  if (method === 'constructor') return
+  Chain.prototype[method] = function (...args) {
+    if (!this._promise) {
+      if (this._hasInvoked === true) {
+        throwIfAlreadyInvoked()
```
mcollina: missing return

index.js
fralonra: Em...I wonder is there any particular reason that we should add a return statement after `throw`.

mcollina: ok

Eomm:
Wasn't this feature a major since:

```
const prom = fastify.inject('/hello');
```

now require:
```
await prom.end()
```
?

mcollina:
No, `await fastify.inject(‘/hello’)` would work as expected.

PR #54, opened 2019-10-10
Add json() handler for Response

fralonra:
Add `json()` handler for Response.
And I'm wondering if error thrown by `JSON.stringify` shoud be handled.

#### Checklist

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -77,6 +77,7 @@ declare namespace LightMyRequest {
     payload: string
     body: string
     addTrailers: (trailers: { [key: string]: string }) => void
+    json: (json: object) => void
   }
 }
 
--- a/lib/response.js
+++ b/lib/response.js
@@ -79,6 +79,11 @@ Response.prototype.end = function (data, encoding, callback) {
   this.emit('finish')
 }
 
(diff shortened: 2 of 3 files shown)


fralonra:
> I meant something different. I would like to `json()` in
> 
> https://github.com/fastify/light-my-request/blob/502878ae4fd45be41b4cf690baf212e0a41f38fc/lib/response.js#L92
> 
> . In the response object we get from `light-my-request`.
> Sorry for the confusion!

So you meant a handler that parses the JSON payload of server response?

mcollina:
Supporting something like this would be extremely nice:

```js
const http = require('http')
const inject = require('light-my-request')

const dispatch = function (req, res) {
  const reply = 'Hello World'
  res.writeHead(200, { 'Content-Type': 'text/plain', 'Content-Length': reply.length })
  res.end(reply)
}

const server = http.createServer(dispatch)

inject(dispatch, { method: 'get', url: '/' }, (err, res) => {
  console.log(res.json()) // validates that the content-type is set to JSON and parses it, otherwise it throws
})
```

mcollina:
Something similar to what `fetch` provides

fralonra:
I see. Close then.

PR #55, opened 2019-10-13
Add json() handler for Response

fralonra:
#### Checklist

### DIFF
--- a/lib/response.js
+++ b/lib/response.js
@@ -109,7 +109,12 @@ function generatePayload (response) {
   res.trailers = response._lightMyRequest.trailers
 
   // Prepare payload parsers
-  res.json = parseJsonPayload.bind(res)
+  res.json = function parseJsonPayload () {
+    if (this.headers['content-type'] !== 'application/json') {
+      throw new Error('The content-type of the response if not application/json')
+    }
+    return JSON.parse(this.payload)
+  }
 
   return res
 }
@@ -121,13 +126,6 @@ function getNullSocket () {
       setImmediate(callback)
     }
   })
-}
-
-function parseJsonPayload () {
-  if (this.headers['content-type'] !== 'application/json') {
-    throw new Error('The content-type of the response if not application/json')
-  }
-  return JSON.parse(this.payload)
 }
 
 module.exports = Response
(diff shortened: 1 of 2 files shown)


lib/response.js
```diff
@@ -120,4 +123,15 @@ function getNullSocket () {
...
 
+function parseJsonPayload () {
+  if (this.headers['content-type'] !== 'application/json') {
+    throw new Error('The content-type of the response if not application/json')
+  }
+  try {
+    return JSON.parse(this.payload)
+  } catch (e) {
+    throw new Error(e)
+  }
```
SerayaEryn: Why do you catch and rethrow the error?

fralonra: My fault. Fixed!

lib/response.js
```diff
@@ -108,6 +108,9 @@ function generatePayload (response) {
   res.body = res.payload
   res.trailers = response._lightMyRequest.trailers
 
+  // Prepare payload parsers
+  res.json = parseJsonPayload.bind(res)
```
mcollina: I would not bind this, but rather attach it to `res` directly.

fralonra: Done

lib/response.js
```diff
@@ -108,6 +108,14 @@ function generatePayload (response) {
   res.body = res.payload
   res.trailers = response._lightMyRequest.trailers
 
+  // Prepare payload parsers
+  res.json = function parseJsonPayload () {
+    if (this.headers['content-type'] !== 'application/json') {
```
mcollina: You should use `indexOf` instead. The most common form is `application/json; charset=utf-8` because... some browser compatibility issue :/.

fralonra: Done

lib/response.js
```diff
@@ -108,6 +108,14 @@ function generatePayload (response) {
   res.body = res.payload
   res.trailers = response._lightMyRequest.trailers
 
+  // Prepare payload parsers
+  res.json = function parseJsonPayload () {
+    if (this.headers['content-type'].indexOf('application/json') < 0) {
+      throw new Error('The content-type of the response if not application/json')
```
StarpTech: Typo `is not application/json`

fralonra: Fixed

PR #62, opened 2019-10-27
fix regression introduced in v3.6.0 with string urls

mcollina:
#### Checklist

### DIFF
--- a/index.js
+++ b/index.js
@@ -99,6 +99,9 @@ function doInject (dispatchFunc, options, callback) {
 }
 
 function Chain (dispatch, option) {
+  if (typeof option === 'string') {
+    option = { url: option }
+  }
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -99,6 +99,9 @@ function doInject (dispatchFunc, options, callback) {
 }
 
 function Chain (dispatch, option) {
+  if (typeof option === 'string') {
+    option = { url: option }
+  }
```
fralonra: Can we just assign `this.option` here and wrap the next line in a `else` block? Because if `option` is a string, there is no need to call `Object.assign`.

StarpTech: With `object.assign` you create a shallow copy of the options to prevent side effects that's fine.

fralonra:
LGTM

PR #67, opened 2019-11-18
feat: freeze proto to work with express

salmanm:
This package doesn't work with express because express replaces prototype with it's own version. 
[This](https://github.com/expressjs/express/blob/master/lib/middleware/init.js#L36) is where it happens.

We however need our methods to remain intact after express initializes to be able to intercept writes and record them.

To address this I've raised this PR that freezes our 5 prototype methods by making them non-writable so that express can not replace them.

I'm not sure if this approach is good, so haven't yet updated the docs, if you think it's ok to go this way, I'll update the docs to show how it can be used with express.

#### Checklist


salmanm:
Not sure how applying it to prototype directly would work. A non writable prototype won't stay like that when an instance is created. Or am I missing something? Let me know how you mean.

Yes will add more tests once the approach is finalized.

index.js
```diff
@@ -74,6 +75,9 @@ function inject (dispatchFunc, options, callback) {
 function doInject (dispatchFunc, options, callback) {
   options = (typeof options === 'string' ? { url: options } : options)
 
+  let express = false;
+  ({ express, ...options } = options)
```
mcollina: This is nice to read, but it's slower than just reading a property.

salmanm: Just wanted to delete it from `options` as it's not needed further. Would you rather just keep it?

mcollina: please, otherwise it will reallocate options

salmanm:
express doesn't override any of those methods. It simply `Object.create(http.ServerResponse.prototype)` so our 5 methods are replaced by their original/default versions.

index.js
salmanm: Aye, good point. Done.

index.js
```diff
@@ -88,12 +89,22 @@ function doInject (dispatchFunc, options, callback) {
     const req = new Request(options)
     const res = new Response(req, callback)
 
+    if (options.express) {
```
mcollina: can you extract this into a const at the top.

test/express.test.js
```diff
@@ -0,0 +1,65 @@
...
+
+const test = require('tap').test
+const express = require('express')
+
+const inject = require('../index')
+
+test('works with express with callback', (t) => {
+  t.plan(8)
+
+  const app = express()
```
mcollina: can we potentially automatically detect this?

salmanm: Well, `app` contains a bunch of express-specific properties, so we can probably check for some of them, but not sure how reliable it would be. 

`app.handle`
`app.lazyrouter`
`app.request`
`app.response`

are some of them, do you think they're good enough for a guess?

mcollina: probably not. are app.request and response the prototypes of the req and res? We might be able to auto detect with these.

salmanm: Yep
```js
app.request instanceof http.IncomingMessage === true
app.response instanceof http.ServerResponse === true
```

salmanm: Do you think these two are enough to decide if it's express?
I was leaning towards `inject.express(app, '/path')`

mcollina: Let's do `inject.express(app, '/path')`

mcollina: No need, code is ok

index.js
```diff
@@ -5,6 +5,7 @@ const http = require('http')
 const Ajv = require('ajv')
 const Request = require('./lib/request')
 const Response = require('./lib/response')
+const fixProto = require('./util')
```
RafaelGSS:
```suggestion
const freezePrototype = require('./util')
```
I don't think that this name explain what this method does. Maybe freeze can explain more. What do you think?

index.js
```diff
@@ -82,18 +83,29 @@ function doInject (dispatchFunc, options, callback) {
     }
   }
 
+  const express = options.express
```
RafaelGSS: if instead of defining an express option, we define something with a more generic name?  (in case of another framework that needs it).

index.js
salmanm: Hah, I used that name at first, don't know why I changed it to `fixProto` 😅 
Anyway renamed, thanks.

salmanm:
Aye, doc updated, PR ready for another look.

index.js
salmanm: Feel free to suggest what name you think is more appropriate.

salmanm: No sorry I didn't. 
I don't see how we're disabling request injection here! The aim is to enable request injection "even for express(-like) frameworks".

And since we don't know of any other framework that does the same thing like express (replace req/res proto), I would prefer keeping the flag named as `express`.

RafaelGSS: Good point! I don't know any other framework that does the same thing too. So, express sounds good for now. Maybe in the future, we can change.

salmanm:
@mcollina Are we gonna merge this one soon OR in the next planned release?

mcollina:
I was trying to release this and the tests are failing locally for me (Linux) with:

```
TypeError: Cannot read property 'length' of null
    at Request.Readable.unpipe (_stream_readable.js:852:19)
    at unpipe (/home/matteo/repositories/light-my-request/node_modules/unpipe/index.js:47:12)
    at send (/home/matteo/repositories/light-my-request/node_modules/finalhandler/index.js:306:3)
    at /home/matteo/repositories/light-my-request/node_modules/finalhandler/index.js:133:5
    at /home/matteo/repositories/light-my-request/node_modules/express/lib/router/index.js:635:15
    at next (/home/matteo/repositories/light-my-request/node_modules/express/lib/router/index.js:260:14)
    at expressInit (/home/matteo/repositories/light-my-request/node_modules/express/lib/middleware/init.js:40:5)
    at Layer.handle [as handle_request] (/home/matteo/repositories/light-my-request/node_modules/express/lib/router/layer.js:95:5)
    at trim_prefix (/home/matteo/repositories/light-my-request/node_modules/express/lib/router/index.js:317:13)
    at /home/matteo/repositories/light-my-request/node_modules/express/lib/router/index.js:284:7
1..2
```

mcollina:
I'll revert the commit. Would you mind to send a fresh PR?

salmanm:
Oh, yes sure, I'll look into it.

PR #69, opened 2019-12-11
Override Response.setTimeout to make response timeout handler work

gyszalai:
#### Checklist

### DIFF
--- a/lib/response.js
+++ b/lib/response.js
@@ -48,11 +48,9 @@ util.inherits(Response, http.ServerResponse)
 
 Response.prototype.setTimeout = function (msecs, callback) {
   this.timeoutHandle = setTimeout(() => {
-    this.socket.emit('timeout')
-    if (callback) {
-      callback()
-    }
+    this.emit('timeout')
   }, msecs)
+  this.on('timeout', callback)
   return this
 }
 
--- a/lib/response.js
+++ b/lib/response.js
@@ -46,18 +46,17 @@ function Response (req, onEnd, reject) {
 
 util.inherits(Response, http.ServerResponse)
 
-let timeoutHandle
 Response.prototype.setTimeout = function (msecs, callback) {
-  if (callback) {
-    timeoutHandle = setTimeout(callback, msecs)
-  }
+  this.timeoutHandle = setTimeout(() => {
+    this.socket.emit('timeout')
+    if (callback) {
+      callback()
+    }
+  }, msecs)
   return this
 }
 
 Response.prototype.writeHead = function () {
-  if (timeoutHandle) {
-    clearTimeout(timeoutHandle)
-  }
   const result = http.ServerResponse.prototype.writeHead.apply(this, arguments)
 
   const _headers = this.getHeaders ? this.getHeaders() : this._headers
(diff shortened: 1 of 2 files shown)


lib/response.js
```diff
@@ -46,7 +46,18 @@ function Response (req, onEnd, reject) {
 
 util.inherits(Response, http.ServerResponse)
 
+let timeoutHandle
+Response.prototype.setTimeout = function (msecs, callback) {
+  if (callback) {
+    timeoutHandle = setTimeout(callback, msecs)
```
mcollina: We will need to emit the `'timeout'` event when the setTimeout triggers (independently of the presence of the callback)

See https://nodejs.org/api/net.html#net_socket_settimeout_timeout_callback for more details.

lib/response.js
```diff
@@ -46,7 +46,18 @@ function Response (req, onEnd, reject) {
 
 util.inherits(Response, http.ServerResponse)
 
+let timeoutHandle
```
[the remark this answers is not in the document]
gyszalai: @Eomm Id don't really get it. do you mean I should set it as a property and access it as `this.timeoutHandle`?

lib/response.js
```diff
@@ -46,7 +46,18 @@ function Response (req, onEnd, reject) {
...
+Response.prototype.setTimeout = function (msecs, callback) {
+  if (callback) {
```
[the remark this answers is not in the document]
gyszalai: ok, shall I throw an error if no callback is specified?

lib/response.js
gyszalai: @mcollina according to the docs, the timeout event should be emitted by the socket. how can I cet a reference to the underlying socket?

gyszalai: is calling `this.socket.emit('timeout')` enough?

lib/response.js
```diff
@@ -46,7 +46,18 @@ function Response (req, onEnd, reject) {
...
+    timeoutHandle = setTimeout(callback, msecs)
+  }
+  return this
+}
+
 Response.prototype.writeHead = function () {
+  if (timeoutHandle) {
+    clearTimeout(timeoutHandle)
```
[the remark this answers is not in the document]
gyszalai: @Eomm ok, I'll move it there

lib/response.js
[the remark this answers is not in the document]
gyszalai: According to the http docs, the callback is optional: https://nodejs.org/api/http.html#http_response_settimeout_msecs_callback

lib/response.js
```diff
@@ -46,6 +46,16 @@ function Response (req, onEnd, reject) {
 
 util.inherits(Response, http.ServerResponse)
 
+Response.prototype.setTimeout = function (msecs, callback) {
+  this.timeoutHandle = setTimeout(() => {
+    this.socket.emit('timeout')
```
mcollina: It should be `this.emit('timeout')`.

lib/response.js
```diff
@@ -46,6 +46,16 @@ function Response (req, onEnd, reject) {
...
+    if (callback) {
+      callback()
+    }
```
mcollina: I would replace this with `this.on('timeout', callback)` outside of the setTimeout callback

gyszalai: done

lib/response.js
gyszalai: done

gyszalai:
Somebody please, merge this PR.

PR #72, opened 2020-01-03
Easy cookie

### LINKED ISSUE #71: inject cookie
## 🚀 Feature Proposal

Add the feature to send a request with cookies and read the cookies in the response easily

## Motivation

Simplify the cookie usage

## Example

```js
inject(dispatch, { method: 'get', url: '/', cookies: {
 foo: 'bar'
} }, (err, res) => {
  console.log(res.payload)
  console.log(res.cookies()) // returned by the server (same usage of .json() feature)
})
```

Eomm:
closes #71

### DIFF
--- a/lib/request.js
+++ b/lib/request.js
@@ -54,7 +54,11 @@ function Request (options) {
 
   if (options.cookies) {
     const { cookies } = options
-    this.headers.cookie = Object.keys(cookies).map(key => cookie.serialize(key, cookies[key])).join('; ')
+    const cookieValues = Object.keys(cookies).map(key => cookie.serialize(key, cookies[key]))
+    if (this.headers.cookie) {
+      cookieValues.unshift(this.headers.cookie)
+    }
+    this.headers.cookie = cookieValues.join('; ')
   }
 
   this.connection = {
--- a/lib/request.js
+++ b/lib/request.js
@@ -4,6 +4,7 @@
 
 const { Readable } = require('readable-stream')
 const util = require('util')
+const cookie = require('cookie')
 
 const parseURL = require('./parseURL')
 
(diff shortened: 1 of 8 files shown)


lib/request.js
```diff
@@ -50,6 +52,11 @@ function Request (options) {
   this.headers['user-agent'] = this.headers['user-agent'] || 'lightMyRequest'
   this.headers.host = this.headers.host || options.authority || hostHeaderFromURL(parsedURL)
 
+  if (options.cookies) {
+    const { cookies } = options
+    this.headers.cookie = Object.keys(cookies).map(key => cookie.serialize(key, cookies[key])).join('; ')
```
mcollina: I think we should document the fact that the cookies option take precedence over the cookie header.

lib/response.js
```diff
@@ -127,6 +129,10 @@ function generatePayload (response) {
     return JSON.parse(this.payload)
   }
 
+  res.cookies = function parseCookies () {
+    return setCookie.parse(this)
+  }
```
mcollina: I would make this a getter instead. I would also create a new class that inherits from ServerResponse with it. (use util.inherits, not class).

lib/request.js
Eomm: Maybe we could just append, so the user will decide

lib/response.js
Eomm: Is not the inherit already done here?

https://github.com/fastify/light-my-request/blob/7acd5444048b1bad60bca6edd163401b51e0181f/lib/response.js#L47

Gotcha for the getter

Eomm:
Good point, but I found [this messge](https://github.com/jshttp/cookie/issues/58#issuecomment-230946655) by the creator of `cookie` that says that the module's parse function is only designed to parse the Cookie header, not the Set-Cookie header.

mcollina:
@Eomm would you mind to send a PR to update the fastify docs?

PR #73, opened 2020-01-31
Validate headers value

davidebianchi:
Headers value must be a string, but adding key to a javascript object sometimes by mistake you could add a number, an array, an object or something different from a string.
In this case, the test may incorrectly pass.

#### Checklist

### DIFF
--- a/index.js
+++ b/index.js
@@ -37,9 +37,7 @@ const schema = {
     },
     headers: {
       type: 'object',
-      additionalProperties: {
-        type: 'string'
-      }
+      additionalProperties: true
     },
     query: {
       type: 'object',
--- a/index.js
+++ b/index.js
@@ -37,7 +37,9 @@ const schema = {
     },
     headers: {
       type: 'object',
-      additionalProperties: true
+      additionalProperties: {
+        type: 'string'
+      }
     },
     query: {
       type: 'object',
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -37,7 +37,9 @@ const schema = {
     },
     headers: {
       type: 'object',
-      additionalProperties: true
+      additionalProperties: {
+        type: 'string'
```
Eomm: I'm not sure we should throw an error to users that set a number or simply `null`
We should serialize the object in the `lib/request.js` instead because, for example, a `date` object will be the ISO string

davidebianchi: Hi @Eomm, thanks for the review!
I thought about it, but serialize the input headers object could bring to other unexpected behaviour. For example, how we should serialize the `null` value, to string `"null"` or to empty string or should we remove the key? And a date object, should we serialize to iso string or to unix timestamp, or another format?
Being this library for testing and debugging purpose, I think users should have full control of the input, warning them in case of incorrect use

davidebianchi:
I checked how Node.js core http library works for request and implemented expected behavior.
Headers are all converted to stringified value, except for type `undefined` that throw an invalid header error.

delvedor:
Landed in [`v3.7.1`](https://github.com/fastify/light-my-request/releases/tag/v3.7.1)!

PR #76, opened 2020-03-23
fix incorrect typescript defs for query

rburgst:
this PR fixes wrong typescript definitions for the `query` object in `HttpInjectOptions`.

#### Checklist

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -34,7 +34,7 @@ declare namespace LightMyRequest {
       protocal?: string
       hostname?: string
       port?: string | number
-      query?: string
+      query?: string | any
     }
     path?: string | {
       pathname: string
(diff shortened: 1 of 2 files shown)


index.d.ts
```diff
@@ -34,7 +34,7 @@ declare namespace LightMyRequest {
       protocal?: string
       hostname?: string
       port?: string | number
-      query?: string
+      query?: string | any
```
mcollina: Can we say it's an object with unknown properties?

index.d.ts
```diff
@@ -44,7 +44,7 @@ declare namespace LightMyRequest {
       query?: string
     }
     headers?: http.IncomingHttpHeaders | http.OutgoingHttpHeaders
-    query?: string
+    query?: string | any
```
Eomm:
```suggestion
    query?: string | { [k: string]: string }
```

rburgst:
when can I expect a new release which includes this?

mcollina:
done now

PR #80, opened 2020-03-28
move to dependabot

mcollina:
as titled


mcollina:
It did not cause issues so far, I won't remove it.

PR #81, opened 2020-04-04
fix(): type in protocol

chyzwar:
Fix typo in types.


chyzwar:
added ts test. Are you guys considering moving code to TS or only bundle definitions?

mcollina:
We are not going to move this to ts.

Issue #82, opened 2020-04-07
Type definition for Response.json() is inaccurate


robcresswell:
## 🐛 Bug Report

The type definition for the Response interface has a `.json()` function, which returns `object`. `object` in TS is the same as `{}`, which is not what the json() response value is.

Personally I'd suggest using something like the `Json` example here: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#more-recursive-type-aliases which is 

```ts
type Json =
    | string
    | number
    | boolean
    | null
    | { [property: string]: Json }
    | Json[];
```

Another option would be to copy Node's own `JSON.parse()` return which is `any`, or even to return `unknown`.

## To Reproduce

I personally stumbled across then when using `fastify.inject`, and then attempted to destructure a key from the `.json()` response.

```ts
const createResult = server.inject({ method: 'POST', url: '/create'});
const { id } = createResult.json();
```

This will error, as it "correctly" tells me that `id` is not an attribute on `{}`.

## Expected behavior

I'd expect this to be more correct, or at least more loosely typed. As it stands, its strict and wrong, I believe.

## Your Environment

- *node version*: 13
- *fastify version*: >=2.0.0
- *os*: Linux

mcollina:
Would you like to send a PR to fix this?

robcresswell:
Gladly! Do you have a preference on the solution? Changing it to `any` is probably the easiest from a maintainers point of view, and matches with the `JSON.parse()` return value IIRC.

mcollina:
I’m totally fine with the result of JSON.parse()

robcresswell:
PR here: 

https://github.com/fastify/light-my-request/pull/83

PR #83, opened 2020-04-10
fix: Match Response.json() type to JSON.parse()

robcresswell:
This patch updates the `Response.json()` type to mirror the result of Node's `JSON.parse()` interface. Previously, `json()` was defined as returning `object`; while this may appear correct from a "human" point-of-view, this is actually the same as saying that the `json()` always returns an empty Object, which is incorrect.

Fixes: #82

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -79,7 +79,7 @@ declare namespace LightMyRequest {
     trailers: { [key: string]: string }
     payload: string
     body: string
-    json: () => object
+    json: () => any
     cookies: Array<object>
   }
 


index.d.ts
```diff
@@ -79,7 +79,7 @@ declare namespace LightMyRequest {
     trailers: { [key: string]: string }
     payload: string
     body: string
-    json: () => object
+    json: () => any
```
Eomm, line 82: I was asking myself how [`@types/node`](https://github.com/DefinitelyTyped/DefinitelyTyped) has defined what `JSON.parse` returns but I didn't find anything useful.
Do you think their definition could help to reduce the `any` range?

robcresswell: There's a bit of discussion on this in #82 which might be of interest

robcresswell:
Definitely! Sorry, had no idea type defs like this could be tested; I've only worked with pure TS so far, rather than JS + defs. Will get that done this weekend.

kibertoad:
I believe this was superseded by #86

Eomm:
Yes it is 👌

Closing

PR #84, opened 2020-04-18
Add overload for `inject` to support builder-style usage

kibertoad:
fixes https://github.com/fastify/fastify/issues/2142

#### Checklist

- [X] documentation is changed or added
- [X] commit message and code follows [Code of conduct](https://github.com/fastify/fastify/blob/master/CODE_OF_CONDUCT.md)

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -14,6 +14,7 @@ declare namespace LightMyRequest {
     dispatchFunc: DispatchFunc,
     options?: string | InjectOptions
   ): Chain
+  function inject (): Chain
   function inject (
     dispatchFunc: DispatchFunc,
     options: string | InjectOptions,
--- a/test/index.test-d.ts
+++ b/test/index.test-d.ts
@@ -73,7 +73,7 @@ inject(dispatch)
 
 expectType<Chain>(inject(dispatch))
 expectType<Promise<Response>>(inject(dispatch).end())
-expectType<Chain | Promise<Response>>(
+expectType<Promise<Response>>(
     inject()
         .get('/')
         .headers({ foo: 'bar' })
@@ -91,4 +91,3 @@ type ParsedValue = { field: string }
 const response: Response = await inject(dispatch)
 const parsedValue: ParsedValue = response.json()
 expectType<ParsedValue>(parsedValue)
-
--- a/test/index.test-d.ts
+++ b/test/index.test-d.ts
@@ -71,7 +71,9 @@ inject(dispatch)
     console.log(res.payload)
   })
 
+expectType<Chain>(inject())
 expectType<Chain>(inject(dispatch))
 expectType<Chain>(inject(dispatch, { method: 'get', url: '/' }))
 // @ts-ignore tsd supports top-level await, but normal ts does not so ignore
 expectType<Response>(await inject(dispatch, { method: 'get', url: '/' }))
+
--- a/test/index.test-d.ts
+++ b/test/index.test-d.ts
@@ -71,8 +71,15 @@ inject(dispatch)
     console.log(res.payload)
   })
 
-expectType<Chain>(inject())
-expectType<Chain>(inject(dispatch))
+expectType<Chain | Promise<Response>>(
+    inject()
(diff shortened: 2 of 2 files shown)


index.d.ts
```diff
@@ -14,6 +14,7 @@ declare namespace LightMyRequest {
     dispatchFunc: DispatchFunc,
     options?: string | InjectOptions
   ): Chain
+  function inject (): Chain
```
Eomm, line 17: This is wrong and it throws:

> AssertionError [ERR_ASSERTION]: dispatchFunc should be a function

kibertoad: @Eomm This is strange, because:

1) This is how I use it in my code and it works:
```
    const response = await app.inject().get('/health')
```

2) This is how code in question looks like:
```
function inject (dispatchFunc, options, callback) {
  if (typeof callback === 'undefined') {
    return new Chain(dispatchFunc, options)
  } else {
    return doInject(dispatchFunc, options, callback)
  }
}

function Chain (dispatch, option) {
  if (typeof option === 'string') {
    this.option = { url: option }
  } else {
    this.option = Object.assign({}, option)
  }

  this.dispatch = dispatch
  this._hasInvoked = false
  this._promise = null

  if (this.option.autoStart !== false) {
    process.nextTick(() => {
      if (!this._hasInvoked) {
        this.end()
      }
    })
  }
}
```

Why would that throw and not just return a Chain?

kibertoad: I assume you are trying to `await` the chain without passing any `dispatchFunc` prior to that. That is not the intention of this particular overload.

Eomm: app is a fastify instance, light my request doesn't know it

```
const inject = require('light-my-request')

// chaining methods
inject()
  .get('/') // set the request method to GET, and request URL to '/'
  .headers({ foo: 'bar' }) // set the request headers
  .query({ foo: 'bar' }) // set the query parameters
  .end((err, res) => {
    console.log(res.payload)
  })
```

kibertoad: @Eomm That's my point. In your example calling just `inject()` should be perfectly fine, it just returns a builder on which you keep calling `.get()` etc.
`app` is not relevant in this case.

kibertoad: I can change my test to use the whole sequence of
```
inject()
  .get('/') // set the request method to GET, and request URL to '/'
  .headers({ foo: 'bar' }) // set the request headers
  .query({ foo: 'bar' }) // set the query parameters
  .end((err, res) => {
    console.log(res.payload)
  })
```
if you prefer that. Currently even that is not supported by typings.

Eomm: > if you prefer that

It is not matter of preferences, that kind of usage it is not supported and it should not permitted in typings

kibertoad: Okay, I adjusted test accordingly

kibertoad:
@Eomm Is anything missing?

Eomm:
@mcollina could you recheck?

I think this PR adds a wrong kind of usage for light my request since `inject()` (without any arguments) will throw an exception later and the user can't sent the dispatch function with the `Chain` object.

Here an example that throws, but it will be permitted:

```
const inject = require('light-my-request')

// chaining methods
inject()
  .get('/') // set the request method to GET, and request URL to '/'
  .headers({ foo: 'bar' }) // set the request headers
  .query({ foo: 'bar' }) // set the query parameters
  .end((err, res) => {
    console.log(res.payload)
  })
```

mcollina:
@Eomm I don’t know, I trust your judgment :(. I currently do not have many cycles to help diagnose minor TS incompatibilities :(.

kibertoad:
Closing as it seems to be incorrect usage.

Eomm:
I think your issue @kibertoad  has been fixed in https://github.com/fastify/fastify/pull/2209

You can verify it changing your dependency to:

```
"dependencies": {
  "fastify": "git://github.com/fastify/fastify.git#394d0d4cdcea3d9f98a288dfc816ca31eec62605"
}
```

PR #86, opened 2020-04-18
Fix type for Response.json()

### LINKED ISSUE #82: Type definition for Response.json() is inaccurate

kibertoad:
fixes #82
inspired by #83

type `any` was chosen as per instructions from @mcollina to replicate what JSON.parse is returning.

#### Checklist

- [X] documentation is changed or added
- [X] commit message and code follows [Code of conduct](https://github.com/fastify/fastify/blob/master/CODE_OF_CONDUCT.md)

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -79,7 +79,7 @@ declare namespace LightMyRequest {
     trailers: { [key: string]: string }
     payload: string
     body: string
-    json: () => object
+    json: () => any
     cookies: Array<object>
   }
 
--- a/test/index.test-d.ts
+++ b/test/index.test-d.ts
@@ -75,3 +75,9 @@ expectType<Chain>(inject(dispatch))
 expectType<Chain>(inject(dispatch, { method: 'get', url: '/' }))
 // @ts-ignore tsd supports top-level await, but normal ts does not so ignore
 expectType<Response>(await inject(dispatch, { method: 'get', url: '/' }))
+
+type ParsedValue = { field: string }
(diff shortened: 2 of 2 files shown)


robcresswell:
Thanks! Sorry to let you down on getting that done

kibertoad:
@robcresswell You didn't! I finished what you started, without your PR there wouldn't be mine, so all the credit goes to you.

PR #88, opened 2020-04-27
add form option

Eomm:
I would like to introduce this option to easy the `form` usage when `inject` with fastify

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,8 +87,7 @@ try {
   const res = await inject(dispatch, {
     method: 'post',
     url: '/upload',
-    payload: form.payload,
-    headers: form.headers
+    ...form
   })
   console.log(res.payload)
 } catch (err) {
--- a/README.md
+++ b/README.md
@@ -72,21 +72,23 @@ try {
 }
 ```
 
-File uploads (multipart/form-data) can be achieved by using [form-data](https://github.com/form-data/form-data) package as shown below:
+File uploads (`multipart/form-data`) or form submit (`x-www-form-urlencoded`) can be achieved by using [form-auto-content](https://github.com/Eomm/form-auto-content) package as shown below:
 
 ```js
-const FormData = require('form-data')
+const formAutoContent = require('form-auto-content')
 const fs = require('fs')
 
 try {
-  const form = new FormData()
-  form.append('myfile', fs.createReadStream(`./path/to/file`))
+  const form = formAutoContent({
+    myField: 'hello',
+    myFile: fs.createReadStream(`./path/to/file`)
+  })
 
   const res = await inject(dispatch, {
     method: 'post',
     url: '/upload',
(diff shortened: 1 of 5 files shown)


Eomm:
Could the solution be to add the form dependency optional in the package json?

I'm targeting to have a fast way to send a form as got does

mcollina:
setting it as optional would still have npm download it :/.

Eomm:
> setting it as optional would still have npm download it :/.

optional by input like it was done in fastify-compress for the brotli support could be fine for you?
So it must be a user action activate this syntactic sugar

example:

```
  const form = await inject(dispatch, {
    method: 'post',
    url: '/upload',
    form: require('form-data'),
    payload: {
      myField: 'hello',
      myfile: fs.createReadStream(`./path/to/file`)
    }
  })
```

mcollina:
that works, yes. However I think this would be slightly better and it does not require any change here, minus a quick addition to the README:

```js
const form = require('simple-form')
const res = await inject(dispatch, {
    method: 'post',
    url: '/upload',
    payload: form({
      myField: 'hello',
      myfile: fs.createReadStream(`./path/to/file`)
    })
  })
```

Eomm:
I didn't find `simple-form` or `form-data` support this syntax:

```
    payload: form({
      myField: 'hello',
      myfile: fs.createReadStream(`./path/to/file`)
    })
```

Are you suggesting a little new module?

mcollina:
Yes i am ;).

Eomm:
> Yes i am ;).

Before I publish it I would really appreciate if you could give it a check

https://github.com/Eomm/form-method#form-method

mcollina:
I would just return an object like `{ payload, headers }`. I would also stick to a stream for the payload, it seems easier for the recipient. You can get one with `Readable.from(string)`.

Also, I'm unsure about the name, why `method`?

Eomm:
>  I would also stick to a stream for the payload, it seems easier for the recipient

Thanks, I agree, uniformed the output 👍 

> why method?

I was thinking to `<form method=POST>`... and it sounds better than `form-enctype` 😅
I found these two names free on npm
Is it really so bad?

mcollina:
It’s unclear :(. I would go with something with simple or auto in the name.

Eomm:
Renamed `form-auto-content` 👍

PR updated

README.md
```diff
@@ -72,21 +72,23 @@ try {
...
+    myFile: fs.createReadStream(`./path/to/file`)
+  })
 
   const res = await inject(dispatch, {
     method: 'post',
     url: '/upload',
-    payload: form,
-    headers: form.getHeaders()
+    payload: form.payload,
+    headers: form.headers
```
mcollina, line 91: I would use `...form` instead, is more idiomatic

PR #99, opened 2020-09-24
fix: add missing request.socket (#87)

### LINKED ISSUE #87: Missing socket
Hi,
I'm writing some test for a Fastify plugin that uses some `koa` handler to manage routes.

The test is failing due this check of `koa` lib:

```
if (this.socket.encrypted) return 'https';
```
https://github.com/koajs/koa/blob/master/lib/request.js#L402

Considered that:
- the final solution would be to implement a native plugin for Fastify
- light my request says "**Does not use a socket connection**.."
- I can use `listen` instead of `inject` to write my tests

Do you think adding a `this.socket = {}` to the light-my-request is a bad idea?
How would you cover this use case?

humphd:
Fixes #87.  This adds the `.socket` property to `request`, mirroring the `.connection` property

#### Checklist

### DIFF
--- a/lib/request.js
+++ b/lib/request.js
@@ -67,6 +67,7 @@ function Request (options) {
   this.connection = {
     remoteAddress: options.remoteAddress || '127.0.0.1'
   }
+  this.socket = this.connection
 
   // we keep both payload and body for compatibility reasons
   var payload = options.payload || options.body || null
--- a/lib/request.js
+++ b/lib/request.js
@@ -6,8 +6,13 @@ const { Readable } = require('readable-stream')
 const util = require('util')
 const cookie = require('cookie')
 const assert = require('assert')
+const semver = require('semver')
+const warning = require('fastify-warning')()
 
 const parseURL = require('./parseURL')
+
+// request.connectin deprecation https://nodejs.org/api/http.html#http_request_connection
+warning.create('FastifyDeprecationLightMyRequest', 'FST_LIGHTMYREQUEST_DEP01', 'You are accessing "request.connection", use "request.socket" instead.')
 
 /**
  * Get hostname:port
@@ -70,7 +75,9 @@ function Request (options) {
 
   Object.defineProperty(this, 'connection', {
     get () {
-      // if node major ≥13 emit a warning with https://github.com/fastify/fastify-warning
+      if (semver.gte(process.versions.node, '13.0.0')) {
+        warning.emit('FST_LIGHTMYREQUEST_DEP01')
+      }
       return this.socket
     },
     configurable: true
--- a/lib/request.js
+++ b/lib/request.js
@@ -6,7 +6,6 @@ const { Readable } = require('readable-stream')
 const util = require('util')
 const cookie = require('cookie')
 const assert = require('assert')
-const semver = require('semver')
 const warning = require('fastify-warning')()
 
 const parseURL = require('./parseURL')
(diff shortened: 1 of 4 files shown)


Eomm (commented):
Could you add a test case?

lib/request.js
```diff
@@ -67,6 +67,7 @@ function Request (options) {
   this.connection = {
     remoteAddress: options.remoteAddress || '127.0.0.1'
   }
+  this.socket = this.connection
```
jsumners, line 70: Should there be some sort of deprecation here? We'd have to follow the upstream deprecation cycle.

climba03003: I think you can add the deprecate warning in the [typing](https://github.com/fastify/light-my-request/blob/master/index.d.ts#L63) to something like this
```typescript
interface Request extends stream.Readable {
...
// @deprecated
connection: object
...
```

delvedor: I agree with @jsumners.
According to the [Node.js docs](https://nodejs.org/api/http.html#http_request_connection), `.connection` has been deprecated in favor of `.socket`.
I think we should do something like this:
```js
this.socket = {
  remoteAddress: options.remoteAddress || '127.0.0.1'
}

Object.defineProperty(Request, 'connection', {
  get () {
    // if node major ≥13 emit a warning with https://github.com/fastify/fastify-warning
    return this.socket
  }
})
```

mcollina: +1!

humphd:
Thanks for all the feedback.  I've tried to make the changes/updates that were requested.  Let me know if I've done it incorrectly, or there are other things I should consider.

lib/request.js
```diff
@@ -64,10 +64,17 @@ function Request (options) {
...
 
-  this.connection = {
+  this.socket = {
     remoteAddress: options.remoteAddress || '127.0.0.1'
   }
-  this.socket = this.connection
+
+  Object.defineProperty(this, 'connection', {
+    get () {
+      // if node major ≥13 emit a warning with https://github.com/fastify/fastify-warning
```
jsumners, line 73: The comment here is not sufficient. An actual test against against `process.version` is needed along with a warning issued through the interface that `fastify-warning` provides.

humphd: Understood.  Maybe something like https://github.com/fastify/fastify-plugin/blob/d3b2ca47404e622f7c474493bb0624bf95a2e4f4/test/esm/index.test.js#L6, and then use fastify-warning to actually log that message?

jsumners: Correct. Fastify itself uses `fastify-warning` to issue deprecation notices. You can learn from there as well.

humphd: Great, thanks @jsumners, this is really helpful.  I'll dig around and update this to do it correctly.

mcollina:
Any update on this @humphd ?

humphd:
I've submitted another fix, which uses `semver` to do the check against the node version, and `fastify-warning` to emit the deprecation warning.  Thanks to @jsumners for pointing out how I should do this.

A couple things I wasn't sure about related to the way I named things in `warning.create()`, specifically the `CODE` and if there is a standard you want to use.  Also, I noticed that in some places you create all the warnings in a single file.  I did it in the `request.js` file, but could move it.

Also, I ran into a small issue with naming on the code, where I had lowercase letters for the plugin name, but everything gets converted to uppercase.  I'll send a PR to the docs for `fastify-warning` to make that more clear.

lib/request.js
```diff
@@ -64,10 +69,20 @@ function Request (options) {
...
   }
 
-  this.connection = {
+  this.socket = {
     remoteAddress: options.remoteAddress || '127.0.0.1'
   }
 
+  Object.defineProperty(this, 'connection', {
+    get () {
+      if (semver.gte(process.versions.node, '13.0.0')) {
```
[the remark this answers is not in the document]
jsumners: Sounds good to me.

humphd:
Updated to remove `semver` and do the warning on `.connection` regardless of version.

jsumners (approved):
LGTM

mcollina (approved):
lgtm

PR #100, opened 2020-10-01
test: :white_check_mark: improve test line coverage

### LINKED ISSUE #90: 100% code coverage
## 🚀 Feature Proposal

This module could easily reach 100% code coverage. There are just 2 lines missing:

```
----------------------|----------|----------|----------|----------|-------------------|
File                  |  % Stmts | % Branch |  % Funcs |  % Lines | Uncovered Line #s |
----------------------|----------|----------|----------|----------|-------------------|
All files             |    99.15 |    97.48 |    98.04 |     99.1 |                   |
 light-my-request     |      100 |      100 |      100 |      100 |                   |
  index.js            |      100 |      100 |      100 |      100 |                   |
 light-my-request/lib |    98.72 |     96.2 |    97.06 |    98.66 |                   |
  parseURL.js         |      100 |      100 |      100 |      100 |                   |
  request.js          |    98.55 |    98.36 |      100 |    98.51 |               121 |
  response.js         |    98.65 |    85.71 |    95.24 |    98.55 |               130 |
----------------------|----------|----------|----------|----------|-------------------|
```

climba03003:
Resolve #90 

For ```request.js``` L121
I do not see a reason the request will be end early without custom alter the ```_lightMyRequest``` object.
So, I decided to alter it in ```dispatch``` in order to reach the the line and pretend it is a invalid alter by user.

For ```response.ts``` L62
If this plugin do not support node 7.7.0 or early, this is never happen for the false condition.
[NodeJS Docs](https://nodejs.org/api/http.html#http_response_getheaders) for ```getHeaders```

#### Checklist

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -1552,3 +1552,15 @@ test('simulate invalid alter _lightMyRequest.isDone without end', (t) => {
     t.equal(replied, false)
   }, 10)
 })
+
+test('no error for response destory', (t) => {
+  const dispatch = function (req, res) {
+    res.destroy()
+  }
+
+  inject(dispatch, { method: 'GET', url: '/' }, (err, res) => {
+    t.error(err)
+  })
+
+  t.end()
+})
--- a/test/test.js
+++ b/test/test.js
@@ -1508,3 +1508,47 @@ test('example with form-auto-content', (t) => {
     t.ok(/--.+\r\nContent-Disposition: form-data; name="myFile"; filename="LICENSE"\r\n.*/.test(res.payload))
   })
 })
+
+test('simulate invalid alter _lightMyRequest.isDone with end', (t) => {
+  t.plan(2)
+  let end = false
+  const dispatch = function (req, res) {
+    req.resume()
+    req._lightMyRequest.isDone = true
+    req.on('end', () => {
+      end = true
+    })
+  }
+
+  let replied = false
+  inject(dispatch, { method: 'GET', url: '/', simulate: { end: true } }, (notHandledErr, res) => {
+    replied = true
+  })
+
+  setTimeout(() => {
+    t.equal(end, true)
+    t.equal(replied, false)
+  }, 10)
+})
+
+test('simulate invalid alter _lightMyRequest.isDone without end', (t) => {
+  t.plan(2)
+  let end = false
+  const dispatch = function (req, res) {
+    req.resume()
+    req._lightMyRequest.isDone = true
+    req.on('end', () => {
+      end = true
+    })
+  }
+
+  let replied = false
+  inject(dispatch, { method: 'GET', url: '/', simulate: { end: false } }, (notHandledErr, res) => {
+    replied = true
+  })
+
+  setTimeout(() => {
+    t.equal(end, false)
+    t.equal(replied, false)
+  }, 10)
+})
--- a/test/test.js
+++ b/test/test.js
@@ -1525,12 +1525,10 @@ test('simulate invalid alter _lightMyRequest.isDone with end', (t) => {
     replied = true
   })
 
-  process.nextTick(() => {
-    setImmediate(() => {
-      t.equal(end, true)
-      t.equal(replied, false)
-    })
-  })
+  setTimeout(() => {
+    t.equal(end, true)
+    t.equal(replied, false)
+  }, 10)
 })
 
 test('simulate invalid alter _lightMyRequest.isDone without end', (t) => {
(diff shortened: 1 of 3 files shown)


test/test.js
```diff
@@ -1508,3 +1508,47 @@ test('example with form-auto-content', (t) => {
...
+
+  let replied = false
+  inject(dispatch, { method: 'GET', url: '/', simulate: { end: true } }, (notHandledErr, res) => {
+    replied = true
+  })
+
+  setTimeout(() => {
+    t.equal(end, true)
+    t.equal(replied, false)
+  }, 10)
```
mcollina, line 1531: Please avoid `setTimeout`, maybe use `setImmediate` or `process.nextTick`?

mcollina (changes requested):
Can you also add `--100` to https://github.com/fastify/light-my-request/blob/7dff875709c5b4d738cdb1cc6200033c7e7e5410/package.json#L25? So that we fail the run if it does not have 100% code coverage.

test/test.js
kibertoad: What's the difference between these options?

climba03003: @kibertoad  their behavior is different. 

@mcollina ```setImmediate``` or ```process.nextTick``` cannot pass the testes, they fire faster than the event.

climba03003:
```--100``` cannot be added, because of branches and functions is not reached to 100%

for the missing [branch](https://github.com/fastify/light-my-request/blob/master/lib/response.js#L62) is explained as the PR description.
useless this module support node 7.7.0 or earlier, it will never be reached

for the missing [function](https://github.com/fastify/light-my-request/blob/master/lib/response.js#L95)
the testing will be just calling the functions that ends.
nothing can be checked.

**Comment by mcollina** [comment]
https://github.com/fastify/light-my-request/blob/7dff875709c5b4d738cdb1cc6200033c7e7e5410/lib/response.js#L62 < just remove the branch.


Can you add a test that calls https://github.com/fastify/light-my-request/blob/master/lib/response.js#L95?

So we can mark this 100% :).

**Comment by climba03003** [comment]
all works done.
can you help me to add the ```hacktobefest``` flag to https://github.com/fastify/light-my-request/issues/90 before merging?
thanks a lot

**Review by mcollina — APPROVED** [review]
lgtm

File: test/test.js (hunk 1 above)
Comment (RafaelGSS) [reply test/test.js:1531]: I tried some possibilities here, and the following snipper works:
```js
  process.nextTick(() => {
    setImmediate(() => {
      t.equal(end, true)
      t.equal(replied, false)
    })
  })
```

Not sure why the only `process.nextTick` not works, maybe an internal behavior on lib.

climba03003: @RafaelGSS combination of ```process.nextTick``` and ```setImmediate``` did the tick on my side too. I will change to your approach. It seems that even combination of two of them is slightly faster than ```setTimeout(fn, 10)```

test/test.js
```diff
@@ -1547,10 +1549,12 @@ test('simulate invalid alter _lightMyRequest.isDone without end', (t) => {
...
-  setTimeout(() => {
-    t.equal(end, false)
-    t.equal(replied, false)
-  }, 10)
+  process.nextTick(() => {
+    setImmediate(() => {
+      t.equal(end, false)
+      t.equal(replied, false)
+    })
+  })
```
mcollina, line 1557: Can you just use one of those?

mcollina (changes requested):

test/test.js
climba03003: Sadly, I cannot. Using either one of them will fail the test, using both of them still faster than ```setTimeout(fn, 10)``` for 20ms

climba03003: ```process.nextTick``` and ```setImmediate``` result vary on machine. I didnt expect it will fail the test (it is success on my computer). I will revert the new commit.

test/test.js
RafaelGSS: Agree

climba03003: I do not know their are ```t.pass()``` or ```t.fail()``` to use.
I will change the testing method and remove ```setTimeout```.

climba03003:
I have removed the usage of ```setTimeout``` and using ```t.pass()``` and ```t.fail()``` to replace it.
The testing result is good to me.

RafaelGSS (approved):
LGTM.

mcollina (approved):
lgtm

PR #101, opened 2020-10-05
Expose a mock socket that is an EventEmitter like a real one would be

airhorns:
Before this change, if a Fastify user attached an `onTimeout` hook to their server, an error would be thrown in test mode when injecting requests with light-my-request

```
TypeError: request.raw.socket.on is not a function

      at Object.routeHandler [as handler] (../../node_modules/fastify/lib/route.js:350:28)
      at Router.lookup (../../node_modules/find-my-way/index.js:356:14)
      at ../../node_modules/light-my-request/index.js:101:38
      at Request.prepare (../../node_modules/light-my-request/lib/request.js:110:12)
      at ../../node_modules/light-my-request/index.js:101:11
      at doInject (../../node_modules/light-my-request/index.js:97:12)
      at Chain.<computed> [as then] (../../node_modules/light-my-request/index.js:187:23)
          at runMicrotasks (<anonymous>)
```

This fixes the socket object to have an `on` function (as well as the rest of the EventEmitter interface) so that things that add timeout handlers can at least add them without erroring. This doesn't implement actual timeout injection, but I think that could be done in userland by emitting events on the socket manually.

Similar to #99.

### DIFF
--- a/lib/request.js
+++ b/lib/request.js
@@ -9,6 +9,7 @@ const assert = require('assert')
 const warning = require('fastify-warning')()
 
 const parseURL = require('./parseURL')
+const { EventEmitter } = require('events')
 
 // request.connectin deprecation https://nodejs.org/api/http.html#http_request_connection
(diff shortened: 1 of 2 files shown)


L2jLiga (approved):
lgtm

PR #102, opened 2020-10-15
fix: fix response.destroy

salmanm:
- Implemented the missing `Response.prototype.destroy` 
- Fixed test

The test was there for destroy, but it never failed because it didn't have `t.plan` so the test didn't make sure the `inject` response callback was ever called.

#### Checklist

### DIFF
--- a/index.js
+++ b/index.js
@@ -75,6 +75,18 @@ function inject (dispatchFunc, options, callback) {
   }
 }
 
+function makeRequest (dispatchFunc, server, req, res) {
+  req.once('error', function (err) {
+    if (this.destroyed) res.destroy(err)
+  })
+
+  req.once('close', function () {
+    if (this.destroyed && !this._error) res.destroy()
+  })
+
+  return req.prepare(() => dispatchFunc.call(server, req, res))
+}
+
 function doInject (dispatchFunc, options, callback) {
   options = (typeof options === 'string' ? { url: options } : options)
 
@@ -88,29 +100,17 @@ function doInject (dispatchFunc, options, callback) {
 
   const server = options.server || {}
 
-  const makeRequest = (req, res) => {
-    req.once('error', function (err) {
-      if (this.destroyed) res.destroy(err)
-    })
-
-    req.once('close', function () {
-      if (this.destroyed && !this._error) res.destroy()
-    })
-
-    return req.prepare(() => dispatchFunc.call(server, req, res))
-  }
-
   if (typeof callback === 'function') {
     const req = new Request(options)
     const res = new Response(req, callback)
 
-    return makeRequest(req, res)
+    return makeRequest(dispatchFunc, server, req, res)
   } else {
     return new Promise((resolve, reject) => {
       const req = new Request(options)
       const res = new Response(req, resolve, reject)
 
-      makeRequest(req, res)
+      makeRequest(dispatchFunc, server, req, res)
     })
   }
 }
--- a/index.js
+++ b/index.js
@@ -88,17 +88,29 @@ function doInject (dispatchFunc, options, callback) {
 
   const server = options.server || {}
 
+  const makeRequest = (req, res) => {
+    req.once('error', function (err) {
+      if (this.destroyed) res.destroy(err)
+    })
+
+    req.once('close', function () {
+      if (this.destroyed && !this._error) res.destroy()
+    })
+
+    return req.prepare(() => dispatchFunc.call(server, req, res))
+  }
+
   if (typeof callback === 'function') {
     const req = new Request(options)
     const res = new Response(req, callback)
 
-    return req.prepare(() => dispatchFunc.call(server, req, res))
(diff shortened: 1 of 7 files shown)


lib/response.js
```diff
@@ -91,7 +91,9 @@ Response.prototype.end = function (data, encoding, callback) {
   this.emit('finish')
 }
 
-Response.prototype.destroy = function () {}
+Response.prototype.destroy = function (err) {
```
mcollina, line 94: destroy can be called without an error, and you should emit only in that case. The previous test was correct.

This is the original implementation (for reference): https://github.com/nodejs/node/blob/f59d4e05a209173406465a26330465413264a443/lib/_http_outgoing.js#L288-L306

salmanm: If destroy is called without an `error`, what do I emit then? How do we terminate the request?
My understanding was that regardless of whether it is called with or without an error, we gotta end the request. and `onEndFailure` handler does this well.

Also, regarding the previous test, are you sure it was correct? I already tried with a breakpoint and control never comes at [this line](https://github.com/fastify/light-my-request/pull/102/files#diff-a561630bb56b82342bc66697aee2ad96efddcbc9d150665abd6fb7ecb7c0ab2fL1576)

mcollina: Please check the behavior of Node.js core. I think it emits `'close'` and closes the response.

lib/response.js
```diff
@@ -21,14 +21,18 @@ function Response (req, onEnd, reject) {
 
   this._promiseCallback = typeof reject === 'function'
 
+  let called = false
   const onEndSuccess = (payload) => {
+    called = true
```
salmanm, line 26: Didn't add `if (called) return` here as `onEndSuccess` is bound to be called only `once` on `finish`. So adding the condition here it's extraneous.

test/test.js
```diff
@@ -1568,13 +1570,75 @@ test('simulate invalid alter _lightMyRequest.isDone without end', (t) => {
...
+    t.equal(res, null)
+  })
+})
+
+test('compatible with eos', (t) => {
+  t.plan(3)
+
+  const dispatch = function (req, res) {
+    eos(res, (err) => {
+      t.ok(err instanceof Error)
```
salmanm, line 1634: eos thinks this `destroyed` request is indeed premature close. However, we're not interested in what the error is, but the fact that this callback is called. Right?

index.js
```diff
@@ -88,17 +88,29 @@ function doInject (dispatchFunc, options, callback) {
...
+    req.once('error', function (err) {
+      if (this.destroyed) res.destroy(err)
+    })
+
+    req.once('close', function () {
+      if (this.destroyed && !this._error) res.destroy()
+    })
+
+    return req.prepare(() => dispatchFunc.call(server, req, res))
+  }
```
mcollina, line 101: Can you move `makeRequest` one level up? There is no need to allocate a new one for every inject.

lib/request.js
```diff
@@ -174,6 +174,14 @@ Request.prototype._read = function (size) {
   })
 }
 
-Request.prototype.destroy = function () {}
+Request.prototype.destroy = function (error) {
+  this.destroyed = true
```
mcollina, line 178: Please add an `if (this.destroyed) return` at the top of the function

lib/response.js
mcollina: Please add a comment for it

lib/response.js
```diff
@@ -91,7 +97,13 @@ Response.prototype.end = function (data, encoding, callback) {
   this.emit('finish')
 }
 
-Response.prototype.destroy = function () {}
+Response.prototype.destroy = function (error) {
```
mcollina, line 100: Please add an `if (this.destroyed) return` at the top of the function

lib/response.js
```diff
@@ -91,7 +97,13 @@ Response.prototype.end = function (data, encoding, callback) {
...
+  if (error) {
+    this.emit('error', error)
+  }
+
+  this.emit('close')
```
mcollina, line 105: please emit those via nextTick

test/test.js
```diff
@@ -7,6 +7,8 @@ const qs = require('querystring')
 const fs = require('fs')
 const zlib = require('zlib')
 const http = require('http')
+const { finished } = require('stream')
+var eos = require('end-of-stream')
```
mcollina, line 11:
```suggestion
const eos = require('end-of-stream')
```

test/test.js
mcollina: It is correct that the error is 'premature close' as there is no error. Add another test that checks that the error is passed through correctly

mcollina:
CI is not passing, code coverage is lower than 100%

mcollina (approved):
lgtm

delvedor (approved):
LGTM

RafaelGSS (approved):
LGTM.

Issue #109, opened 2020-10-30
Omitting .end() is not possible when using TypeScript because of missing types


squareloop1:
## 🐛 Bug Report

The docs are stating we can also omit `.end()` when using the chainable API and it is returning a promise. When using TypeScript this is not possible because the `Chain` interface does not seem to return a Promise. Using `.then()` without `.end()` in TypeScript gives the error `TS2339: Property 'then' does not exist on type 'Chain'.`

## To Reproduce

Use with TS:
```js
fastify.inject().get('/test').then(() => {});
```
`TS2339: Property 'then' does not exist on type 'Chain'.`

## Expected behavior

It should also work with TypeScript without type errors to be able to omit `.end()` when not needed.

## Your Environment

- *node version*: 14
- *fastify version*: >=3.5.1
- *os*: WSL2
- *light-my-request version*: 4.2.1

PR #110, opened 2020-10-30
Fix typescript chain response

RafaelGSS:
ref: #109 

On my local the following assertion is not valid:
```js
Number(process.versions.node[0]) >= 8 // it returns 1 for node v12.18.3
```
So, changed to `semver` check to prevent bugs

#### Checklist

- [X] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/master/CONTRIBUTING.md#developers-certificate-of-origin-11)

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -100,7 +100,8 @@ declare namespace LightMyRequest {
     payload: (payload: InjectPayload) => Chain
     query: (query: object) => Chain
     cookies: (query: object) => Chain
-    end: (callback?: CallbackFunc) => Promise<Response>
+    end(): Promise<Response>
+    end(callback: CallbackFunc): void
   }
 }
 
--- a/index.d.ts
+++ b/index.d.ts
@@ -86,15 +86,15 @@ declare namespace LightMyRequest {
     cookies: Array<object>
   }
 
-  interface Chain {
-    delete: (url: string) => Chain | Promise<Response>
-    get: (url: string) => Chain | Promise<Response>
-    head: (url: string) => Chain | Promise<Response>
-    options: (url: string) => Chain | Promise<Response>
-    patch: (url: string) => Chain | Promise<Response>
-    post: (url: string) => Chain | Promise<Response>
-    put: (url: string) => Chain | Promise<Response>
-    trace: (url: string) => Chain | Promise<Response>
+  interface Chain extends Promise<Response> {
+    delete: (url: string) => Chain
+    get: (url: string) => Chain
+    head: (url: string) => Chain
+    options: (url: string) => Chain
+    patch: (url: string) => Chain
+    post: (url: string) => Chain
+    put: (url: string) => Chain
+    trace: (url: string) => Chain
     body: (body: InjectPayload) => Chain
     headers: (headers: http.IncomingHttpHeaders | http.OutgoingHttpHeaders) => Chain
     payload: (payload: InjectPayload) => Chain
--- a/index.d.ts
+++ b/index.d.ts
@@ -87,14 +87,14 @@ declare namespace LightMyRequest {
   }
 
   interface Chain {
-    delete: (url: string) => Chain
-    get: (url: string) => Chain
-    head: (url: string) => Chain
-    options: (url: string) => Chain
-    patch: (url: string) => Chain
-    post: (url: string) => Chain
-    put: (url: string) => Chain
-    trace: (url: string) => Chain
+    delete: (url: string) => Chain | Promise<Response>
+    get: (url: string) => Chain | Promise<Response>
+    head: (url: string) => Chain | Promise<Response>
+    options: (url: string) => Chain | Promise<Response>
+    patch: (url: string) => Chain | Promise<Response>
+    post: (url: string) => Chain | Promise<Response>
+    put: (url: string) => Chain | Promise<Response>
+    trace: (url: string) => Chain | Promise<Response>
     body: (body: InjectPayload) => Chain
     headers: (headers: http.IncomingHttpHeaders | http.OutgoingHttpHeaders) => Chain
     payload: (payload: InjectPayload) => Chain
--- a/test/async-await.js
+++ b/test/async-await.js
@@ -1,7 +1,7 @@
 'use strict'
 
 function asyncAwaitTest (t, inject) {
-  t.plan(3)
+  t.plan(4)
 
   t.test('basic async await', async t => {
     const dispatch = function (req, res) {
@@ -39,6 +39,20 @@ function asyncAwaitTest (t, inject) {
     try {
       const chain = inject(dispatch).get('http://example.com:8080/hello')
       const res = await chain.end()
+      t.equal(res.payload, 'hello')
+    } catch (err) {
+      t.fail(err)
+    }
+  })
+
+  t.test('chainable api with async await without end()', async t => {
+    const dispatch = function (req, res) {
+      res.writeHead(200, { 'Content-Type': 'text/plain' })
+      res.end('hello')
+    }
+
+    try {
+      const res = await inject(dispatch).get('http://example.com:8080/hello')
       t.equal(res.payload, 'hello')
     } catch (err) {
       t.fail(err)
--- a/test/index.test-d.ts
+++ b/test/index.test-d.ts
@@ -58,13 +58,15 @@ inject(dispatch, { method: 'get', url: '/', query: { name1: ['value1', 'value2']
   expectResponse(res)
 })
 
-inject(dispatch)
-  .get('/')
-  .end((err, res) => {
-    expectType<Error>(err)
-    expectType<Response>(res)
-    console.log(res.payload)
-  })
+expectType<void>(
+  inject(dispatch)
+    .get('/')
+    .end((err, res) => {
+      expectType<Error>(err)
+      expectType<Response>(res)
+      console.log(res.payload)
+    })
+)
 
 inject(dispatch)
   .get('/')
--- a/test/index.test-d.ts
+++ b/test/index.test-d.ts
@@ -66,6 +66,12 @@ inject(dispatch)
     console.log(res.payload)
   })
 
+inject(dispatch)
+  .get('/')
+  .then((value) => {
+    expectType<Response>(value)
+  })
+
 expectType<Chain>(inject(dispatch))
 expectType<Promise<Response>>(inject(dispatch).end())
 expectType<Chain>(inject(dispatch, { method: 'get', url: '/' }))
--- a/test/test.js
+++ b/test/test.js
@@ -9,6 +9,7 @@ const zlib = require('zlib')
 const http = require('http')
 const { finished } = require('stream')
 const eos = require('end-of-stream')
+const semver = require('semver')
 
 const inject = require('../index')
 const parseURL = require('../lib/parseURL')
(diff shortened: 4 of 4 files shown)


squareloop1:
This only adds union types to the HTTP methods. Shouldn't all of the chainable methods have the union type?
Could the corresponding Chain function also be extended to narrow down the union type automatically somehow?

RafaelGSS:
> This only adds union types to the HTTP methods. Shouldn't all of the chainable methods have the union type?

Only HTTP methods can return Promise (following the example docs).

> Could the corresponding Chain function also be extended to narrow down the union type automatically somehow?

Sorry, I don't understand the question here.

---

Also, I talking with @fastify/typescript about this change. The problem is described [here](https://www.typescriptlang.org/play?#code/PTAECUFMDNIJwFygBYBdUAcDOCQHMBLVZAVwCMA6AYwHsBbYaAQy1QOgE9gAbAvNALR0OAuJACOJSK2AYS3bsACMSgAwAoAgDtU8ZlUgRpGGlqyGA3gF91mnXqYHQAYWRNtoC+tCg8kVEgAFCRw3EiscNp4AJSgALwAfKAACnD0BOYAPFBYJmaQSQA+Lm7a3qCQWgAmQbGJKWl0GZDZxqbmCeo26iCgAPIhoFQhYjqgBHQY3JB0lahMbKagNNCgADJ8aACyHFCS0qjjWKBY9P7IUaC8ANaQCOpU3CzHAKIAHkyT067uWuNfMzmxx+Hi8PjkZF4VFAYlQIS0AEEsKl0uYkGQaDRpkwtLYfLQzKg4CQqKgaHBArEwT5QMQMhRYfCkSimuZ4qBmNxzABucrdHx+VCBAD6ITCJyJUWiSBZzVauXaBVAxRBf2pPnYoECdKwDP8TORjWaVPKNJh+rgf1l5j1p24ADdIIFrKAWEYFfloqbQDYzYzLbSLlg+XiKtVKTKjVkcnkOp5vf6rVHILasY7nVZXccY4qvT4bN0WBwtNDoCQS4s-nRfhGGqiWvaaAQqklqQTWENSn84qAtJAAO6gd6fKaQVWU8rtw5iLBKdlMfvuQ5ULsUQWBADkwEgHwBG+iFEqVUp3NAvQAEvBDPA0nBJ+1p9IAEzzxdETu-Nf+Tfb3ej-e8lYQA) and as @zekth mentioned the new solution for usage would be someting like:

```tsx
  const c = chain.get('/example') as Chain;
  const res1 = await c.end();
```

IMO this solution is a breaking change, since users of `light-my-request` are used to use:

```tsx
  const c = chain.get('/example').end();
```

I'll wait for more opinions of this fix the reach the best approach.

squareloop1:
> Only HTTP methods can return Promise (following the example docs).

As I am seeing it from the code it seems like the other methods are also able to return a promise.

> Sorry, I don't understand the question here.

If we do a union type like this: `get: (url: string) => Chain | Promise<Response>` we have to narrow the type down or cast it in the consuming code to tell typescript what exact type we are handling here, else TS will complain about properties that don't exist on one of the types.

> Also, I talking with @fastify/typescript about this change. The problem is described here and as @zekth mentioned the new solution for usage would be someting like:
  const c = chain.get('/example') as Chain;
  const res1 = await c.end();

Isn't this exactly the same behaviour as it is currently? IMHO the best implementation of this for the consuming user would narrow the type down itself, without the user needing to cast it first. Thats why I thought we could maybe add something to the Chain function to better narrow down the returned type.

RafaelGSS:
> If we do a union type like this: get: (url: string) => Chain | Promise<Response> we have to narrow the type down or cast it in the consuming code to tell typescript what exact type we are handling here, else TS will complain about properties that don't exist on one of the types.

Sure, this behavior that we tend to avoid (cast). As you said, the current PR doesn't fix anything yet because we still need cast properties. The goal of this PR is to find the best approach to deal with this behavior. 

> Isn't this exactly the same behaviour as it is currently? IMHO the best implementation of this for the consuming user would narrow the type down itself, without the user needing to cast it first. Thats why I thought we could maybe add something to the Chain function to better narrow down the returned type.

Yes, this is the goal. What do you have in mind to implement it?

zekth:
For reference we faced this kind of problem lately on this PR. Lot of documentation: https://github.com/fastify/fastify/pull/2615

squareloop1:
> Yes, this is the goal. What do you have in mind to implement it?

I have patched my local dependency with overlapping types like this at the moment which works in my local e2e tests:
```TS
interface Chain {
    delete: (url: string) => Chain & Promise<Response>
    get: (url: string) => Chain & Promise<Response>
    head: (url: string) => Chain & Promise<Response>
    options: (url: string) => Chain & Promise<Response>
    patch: (url: string) => Chain & Promise<Response>
    post: (url: string) => Chain & Promise<Response>
    put: (url: string) => Chain & Promise<Response>
    trace: (url: string) => Chain & Promise<Response>
    body: (body: InjectPayload) => Chain & Promise<Response>
    headers: (headers: http.IncomingHttpHeaders | http.OutgoingHttpHeaders) => Chain & Promise<Response>
    payload: (payload: InjectPayload) => Chain & Promise<Response>
    query: (query: object) => Chain & Promise<Response>
    cookies: (query: object) => Chain & Promise<Response>
    end: (callback?: CallbackFunc) => Promise<Response>
  }
```
As all of the methods except `end()` can be either of type `Chain` or of type `Promise<Response>` this is obviously not really strict and may not be a good final solution. But I am not to deep into the light-my-request code yet to come up with a more strict implementation.

fox1t:
> > Yes, this is the goal. What do you have in mind to implement it?
> 
> I have patched my local dependency with overlapping types like this at the moment which works in my local e2e tests:
> 
> ```ts
> interface Chain {
>     delete: (url: string) => Chain & Promise<Response>
>     get: (url: string) => Chain & Promise<Response>
>     head: (url: string) => Chain & Promise<Response>
>     options: (url: string) => Chain & Promise<Response>
>     patch: (url: string) => Chain & Promise<Response>
>     post: (url: string) => Chain & Promise<Response>
>     put: (url: string) => Chain & Promise<Response>
>     trace: (url: string) => Chain & Promise<Response>
>     body: (body: InjectPayload) => Chain & Promise<Response>
>     headers: (headers: http.IncomingHttpHeaders | http.OutgoingHttpHeaders) => Chain & Promise<Response>
>     payload: (payload: InjectPayload) => Chain & Promise<Response>
>     query: (query: object) => Chain & Promise<Response>
>     cookies: (query: object) => Chain & Promise<Response>
>     end: (callback?: CallbackFunc) => Promise<Response>
>   }
> ```
> 
> As all of the methods except `end()` can be either of type `Chain` or of type `Promise<Response>` this is obviously not really strict and may not be a good final solution. But I am not too deep into the light-my-request code yet to come up with a more strict implementation.
> 
> When I type it with union types like `Chain | Promise<Response>` I get type errors from TS as the compiler can't make out what exact type I am handling and therefore I would need to cast it to the corresponding type I am expecting.

We can't use an intersection here because it will change the return type to a promise-like object that has both promise and chain properties and it is wrong.
We need some kind of generics here, that will allow us to make a conditional return type. Do we have any to use?

fox1t:
Ok we should use the 3rd parameter as conditional ReturnType: https://github.com/fastify/light-my-request/blob/e8fd326574c16a7e6e3d476c479bfb29fb96f0b8/index.js#L71

RafaelGSS:
> Ok we should use the 3rd parameter as conditional ReturnType:
> 
> https://github.com/fastify/light-my-request/blob/e8fd326574c16a7e6e3d476c479bfb29fb96f0b8/index.js#L71

Sorry for late. 

The Http methods couldn't use conditional ReturnType because the signature of the method is the same. Do you have any idea to use it here?

mcollina:
CI is failing

RafaelGSS:
> CI is failing

Yes, we are talking about the best approach to this issue.

fox1t:
> > Ok we should use the 3rd parameter as conditional ReturnType:
> > https://github.com/fastify/light-my-request/blob/e8fd326574c16a7e6e3d476c479bfb29fb96f0b8/index.js#L71
> 
> Sorry for late.
> 
> The Http methods couldn't use conditional ReturnType because the signature of the method is the same. Do you have any idea to use it here?

Can you elaborate? I am not sure I understand what do you mean.

fox1t:
OK, now I got it!

Here I am talking about using the consturctor with a generic parameter to know in andvance which interface to return (promise or chain).

RafaelGSS:
> EDIT: After looking better into this i think it would be better to just add the "promiselike" methods to chain, since they are injected anyway in the prototype:

So +1 on union approach?

fox1t:
The methods that are injected are these:
```js
Object.getOwnPropertyNames(Promise.prototype).filter((method) => method !== "constructor") 
// [ "then", "catch", "finally"]
```

fox1t:
Extending the Promise should just work: https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgMIAs6mRAHpEAEwGdkAFKAewFthiIAeAOUsIgCUJiAHSkegHwBvAFDJkbADYRIALmQAKAK5RJ84mCigA5gEpkAXgFpM2AGTkqteg048+gscm0z5y1es079Rk1hDIFhQ0dIx2vPwQAk7oEHCEbipqyBpaIHqGxhj+gZYhNuEOUU6U3GDADokeKV7pPlmmAUFWobZcEY7i3HBgCOhVyanemX7medZh7UXRXZQaA55pGb7ZY8ETbfaRM8jcSmALNUv1o03jrYXbTpqIEIdDdSOrZ+sXU1fiAEasAJ5u34Q-sgAJIgABWEAQYDIcB+kko8ROz1yrwK706yFi8WgxDcWLYUFxyDgIB+SMaKJaaK2GO6cIRCUUdPh8XkoIhUJh9MRTwpzXykxpxXEAEclNAgQoxRL5JRPhywOScvyNpcMQhKJQANbALhuaVQIFyhVKtZUwUdYU4IhuBBwSSST6ILUAfnkqHtjudADElCAECdURbpiIAL4iERgH7cFDsyHQ2EswiGI46ZAAH2QxvjkejKA9DqdCC1vv9KYU0Cg8gAolAqFAADTIKDou7INUQE4AN0owEIEdAkFgt2QLDYHfEonE4igcAA7vIp9OZ3riaSG05ly2RfIST9N8hw1v51yk7vSQf8Tjz-vlxoekpiKhWG2QEpqJ9oAf72BHwBZLhiDgFxFh0A8bmAaRCUXZAAG0tQgIEHm0ABdUD0kPA9mQZdDtAPAEkNqPDlzBYg+HkBgABUUz3AQFBOSiDw1bVdSJABBOtYQYbMoR2cMIw1fgwGQagfmed1GgjERROeAA6MBYhABQFBbIUTlUy1ZOwxEgA

RafaelGSS:
@fox1t Excellent! Seems good for me. I'll update the code tonight and add some tsd tests for these behavior

fox1t (changes requested):
Can you please add the overload for `end` method?
```
end(): Promise<Response>
end(callback: CallbackFunc): void
```

fox1t (approved):
LGTM

zekth (approved):
nice! lgtm

PR #155, opened 2021-08-18
chore: add failing test for send payload

SimenB:
#### Checklist

When using Express' [`res.send`](http://expressjs.com/en/api.html#res.send), `payload` from `light-my-request` is set to an empty string rather then the value passed.

This is just a failing test as I don't know how to fix it. Pointers very much welcome!

From my debugging:

For some reason, https://github.com/fastify/light-my-request/blob/52dd662a177696bc0e92f56e849665c8d854c961/lib/response.js#L83 is not called.

We end up in express here: https://github.com/expressjs/express/blob/06d11755c99fe4c1cddf8b889a687448b568472d/lib/response.js#L221, which calls https://github.com/nodejs/node/blob/7ca38f05a023666274569343f128c5aed81599f3/lib/_http_outgoing.js#L830, but `light-my-request`'s `Reponse.end` is not invoked, so the `payload` is not correct

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -1706,3 +1706,19 @@ test('passes headers when using an express app', (t) => {
     t.equal(res.headers['some-fancy-header'], 'a very cool value')
   })
 })
+
+test("passes payload when using express' send", (t) => {
+  t.plan(3)
+
+  const app = express()
+
+  app.get('/hello', (req, res) => {
+    res.send('some text')
+  })
+
+  inject(app, { method: 'GET', url: 'http://example.com:8080/hello' }, (err, res) => {
+    t.error(err)
+    t.equal(res.headers['content-length'], '9')
+    t.equal(res.payload, 'some text')
+  })
+})


mcollina:
The problem is that Express replaces the prototype of `req` and `res` with its own.

SimenB:
Can you think of some way to avoid or at least handle that case?

mcollina:
I think our best bet is looking into the Express codebase and see what modifications are done. Then inside `inject()` do some fixes after sending the request to the server.

climba03003:
I think the problem is `light-my-request` is a custom `Request` and `Response` object. `express` is creating it's own full `ServerResponse` object in [here](https://github.com/expressjs/express/blob/06d11755c99fe4c1cddf8b889a687448b568472d/lib/response.js#L42).

So, when you change the `prototype` in [here](https://github.com/expressjs/express/blob/06d11755c99fe4c1cddf8b889a687448b568472d/lib/application.js#L99) (maybe wrong place). It override the `.end` method back to the `ServerResponse` default one.

Not sure if my finding is correct.

mcollina:
Yes, exactly @climba03003.

(That prototype modification is one of the fundamental reason Fastify exists).

PR #156, opened 2021-08-22
fixed parseURL and all tests passed for bug: Value of request.url when using .inject differs from a real server #3257

LobRockyl:
#### Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,6 +6,7 @@
   "dependencies": {
     "ajv": "^8.1.0",
     "cookie": "^0.4.0",
+    "fastify": "^3.20.2",
     "fastify-warning": "^0.2.0",
     "readable-stream": "^3.6.0",
     "set-cookie-parser": "^2.4.1"
--- a/package.json
+++ b/package.json
@@ -6,7 +6,6 @@
   "dependencies": {
     "ajv": "^8.1.0",
     "cookie": "^0.4.0",
-    "fastify": "^3.20.2",
     "fastify-warning": "^0.2.0",
     "readable-stream": "^3.6.0",
     "set-cookie-parser": "^2.4.1"
--- a/lib/parseURL.js
+++ b/lib/parseURL.js
@@ -12,6 +12,9 @@ const BASE_URL = 'http://localhost'
  * @return {URL}
  */
 module.exports = function parseURL (url, query) {
+  if ((typeof url === 'string' || Object.prototype.toString.call(url) === '[object String]') && url.startsWith('//')) {
+    url = BASE_URL + url
+  }
   const result = typeof url === 'object'
     ? Object.assign(new URL(BASE_URL), url)
     : new URL(url, BASE_URL)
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -6,6 +6,7 @@
   "dependencies": {
     "ajv": "^8.1.0",
     "cookie": "^0.4.0",
+    "fastify": "^3.20.2",
```
mcollina, line 9: Unfortunately you should not be using fastify here :(. See the other tests, you can do this with a plain Node.js server.

mcollina (commented):
Good work!

LobRockyl:
`

          test('value of request url when using inject should not differ', async (t) => {
          t.plan(1)
          const ap = fastify();
          ap.get("*", function (request, reply) {
            reply.send({ url: request.url });
          });
          const res = await ap.inject({
            method: "GET",
            url: "http://example.com:8080//foo",
          });
          let v=JSON.parse((res.body))  //added this Json parse
        
          t.equal(v.url,'//foo')
          
        })
`

The code above works fine but as you said I removed fastify and did it with express. Now express doesn't have function inject so:
`

            test('value of request url when using inject should not differ', async (t) => {
              t.plan(1)
            
              const app = express()
            
              app.get('*', (req, res) => {
                console.log(req.url)
                res.jsonp({url:req.url})
              })
            
              inject(app, { method: 'GET', url: 'http://example.com:8080//hello' }, (err, res) => {
                console.log(res)
                t.equal(res.body, '')
              })
            })

`

But that gives an error and somehow the res.body is empty. Do you have any idea?

package.json
LobRockyl: `

          test('value of request url when using inject should not differ', async (t) => {
          t.plan(1)
          const ap = fastify();
          ap.get("*", function (request, reply) {
            reply.send({ url: request.url });
          });
          const res = await ap.inject({
            method: "GET",
            url: "http://example.com:8080//foo",
          });
          let v=JSON.parse((res.body))  //added this Json parse
        
          t.equal(v.url,'//foo')
          
        })
`

The code above works fine but as you said I removed fastify and did it with express. Now express doesn't have function inject so:
`

            test('value of request url when using inject should not differ', async (t) => {
              t.plan(1)
            
              const app = express()
            
              app.get('*', (req, res) => {
                console.log(req.url)
                res.jsonp({url:req.url})
              })
            
              inject(app, { method: 'GET', url: 'http://example.com:8080//hello' }, (err, res) => {
                console.log(res)
                t.equal(res.body, '')
              })
            })

`

But that gives an error and somehow the res.body is empty. Do you have any idea?

mcollina: Do not use express, just use http.createServer().

LobRockyl: Hey did that, it's working could you review

mcollina (approved):
lgtm

Eomm (approved):
Fixes https://github.com/fastify/fastify/issues/3257

PR #159, opened 2021-10-05
added rawHeaders to Request

L2jLiga:
I've noticed that Request does not contain rawHeaders property when I was unable to test proxy example from undici repository, see here: https://github.com/nodejs/undici/blob/main/examples/proxy/proxy.js#L9

#### Checklist

### DIFF
--- a/lib/request.js
+++ b/lib/request.js
@@ -63,6 +63,7 @@ function Request (options) {
   this.method = options.method ? options.method.toUpperCase() : 'GET'
 
   this.headers = {}
+  this.rawHeaders = []
   const headers = options.headers || {}
 
   for (const field in headers) {
@@ -105,6 +106,10 @@ function Request (options) {
   // Set the content-length for the corresponding payload if none set
   if (payload && !payloadResume && !this.headers.hasOwnProperty('content-length')) {
     this.headers['content-length'] = (Buffer.isBuffer(payload) ? payload.length : Buffer.byteLength(payload)).toString()
+  }
+
+  for (const header of Object.keys(this.headers)) {
+    this.rawHeaders.push(header, this.headers[header])
   }
 
   // Use _lightMyRequest namespace to avoid collision with Node
--- a/test/test.js
+++ b/test/test.js
@@ -87,15 +87,16 @@ test('passes headers', (t) => {
 })
 
 test('request has rawHeaders', (t) => {
-  t.plan(2)
+  t.plan(3)
   const dispatch = function (req, res) {
-    res.writeHead(200, { 'Content-Type': 'text/plain' })
-    res.end(req.rawHeaders.toString())
+    t.ok(Array.isArray(req.rawHeaders))
+    t.match(req.rawHeaders, ['super', 'duper', 'user-agent', 'lightMyRequest', 'host', 'example.com:8080'])
+    res.writeHead(200)
+    res.end()
   }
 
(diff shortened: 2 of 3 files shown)


test/test.js
```diff
@@ -86,6 +86,19 @@ test('passes headers', (t) => {
...
+test('request has rawHeaders', (t) => {
+  t.plan(2)
+  const dispatch = function (req, res) {
+    res.writeHead(200, { 'Content-Type': 'text/plain' })
+    res.end(req.rawHeaders.toString())
+  }
+
+  inject(dispatch, { method: 'GET', url: 'http://example.com:8080/hello', headers: { Super: 'duper' } }, (err, res) => {
+    t.error(err)
+    t.equal(res.payload, 'super,duper,user-agent,lightMyRequest,host,example.com:8080')
```
Eomm, line 98: Should not it be `Super` capital letter as the example here https://nodejs.org/api/http.html#http_message_rawheaders?

lib/request.js
```diff
@@ -107,6 +108,8 @@ function Request (options) {
     this.headers['content-length'] = (Buffer.isBuffer(payload) ? payload.length : Buffer.byteLength(payload)).toString()
   }
 
+  Object.keys(this.headers).forEach(header => this.rawHeaders.push(header, this.headers[header]))
```
mcollina, line 111: Could you please use a for...of loop instead?

mcollina (commented):
Could we document this property in the README as well?
Does it need type?

test/test.js
climba03003: We parse all the header to lowercase when the users apply.
https://github.com/fastify/light-my-request/blob/master/lib/request.js#L71

L2jLiga: > Maybe we can move the logic of inserting rawHeaders inside that loop.

not really because some headers (like user-agent, host) are set later:
https://github.com/fastify/light-my-request/blob/master/lib/request.js#L74-L75
https://github.com/fastify/light-my-request/blob/master/lib/request.js#L83
https://github.com/fastify/light-my-request/blob/master/lib/request.js#L102
https://github.com/fastify/light-my-request/blob/master/lib/request.js#L107

L2jLiga:
> Does it need type?

I've added it to Request interface - https://github.com/fastify/light-my-request/pull/159/files#diff-7aa4473ede4abd9ec099e87fec67fd57afafaf39e05d493ab4533acc38547eb8R68

> Could we document this property in the README as well?

It seems like no - I didn't found `Request` description in README but property described in [NodeJS docs](https://nodejs.org/api/http.html#http_message_rawheaders)

lib/request.js
L2jLiga: Sure, I will

L2jLiga: done

test/test.js
```diff
@@ -86,6 +86,19 @@ test('passes headers', (t) => {
   })
 })
 
+test('request has rawHeaders', (t) => {
+  t.plan(2)
+  const dispatch = function (req, res) {
+    res.writeHead(200, { 'Content-Type': 'text/plain' })
+    res.end(req.rawHeaders.toString())
```
mcollina, line 93: Instead of returning a value, could you add an assertion here that verifies that `rawHeaders` is an array?

L2jLiga: done

mcollina (approved):
lgtm

Eomm (approved):

PR #185, opened 2022-03-24
fix: use correct object when calling `json`

SimenB:
#### Checklist

There's already a test for this, so I'm not sure what's going on.

Failing test case using fastify:

```js
import Fastify from 'fastify';

const app = Fastify();

const schema = {
  type: 'object',
  properties: { severity: { type: 'string' } },
  required: ['severity'],
};

app.get('/', { schema: { querystring: schema } }, async () => ({ foo: 'bar' }));

const { json } = await app.inject('/');

console.log(json());
```

```
TypeError: Cannot read properties of undefined (reading 'payload')
    at parseJsonPayload (/Users/simen/repos/light-my-request/node_modules/light-my-request/lib/response.js:147:28)
```

The diff in this PR fixes it

### DIFF
--- a/lib/response.js
+++ b/lib/response.js
@@ -144,7 +144,7 @@ function generatePayload (response) {
     if (res.headers['content-type'].indexOf('application/json') < 0) {
       throw new Error('The content-type of the response is not application/json')
     }
-    return JSON.parse(this.payload)
+    return JSON.parse(res.payload)
   }
 
   return res


mcollina (approved):
lgtm

mcollina:
> LGTM. That's weird tho, it should keep the `this` even destructuring, no?

no, destructuring does not bind().

jsumners:
> LGTM. That's weird tho, it should keep the `this` even destructuring, no?

No, the destructure assigns a new object set to the value of the original object's property. So the `this` context is lost.

Issue #189, opened 2022-05-03
Support custom Request types


wilkmaia:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the feature has not already been requested

### 🚀 Feature Proposal

Support for the `request` object (instance of the `Request` (`lib/request.js`) class) to inherit from a custom class besides `stream.Readable`.

### Motivation

Starting from v12.0.9, [next.js](https://github.com/vercel/next.js) validates the request object's type is an `http.IncomingMessage` instance. If it's not, no error is thrown but the request will fail due to an internal type conversion not being made.

This is an issue for using `light-my-request` to test API requests that use `next.js`. Specifically, this is currently blocking bumping `next.js`' version in [fastify-nextjs](https://github.com/fastify/fastify-nextjs) (https://github.com/fastify/fastify-nextjs/pull/514) because tests don't pass with the current structure of `light-my-request`.

### Example

Supporting a new `options` flag, namely `customRequestType` and adding the following snippet in `lib/request.js:124` (`Request` function definition) should do the trick:

```JavaScript
  if (options.customRequestType) {
    util.inherits(this.constructor, options.customRequestType)
  }
```

wilkmaia:
@mcollina in https://github.com/vercel/next.js/releases/tag/v12.0.9.

Specifically in [this commit](https://github.com/vercel/next.js/commit/3e00a81edeb1da513567456813aa529d92d39d60).

The issue is that before they simply set the TypeScript request type to `http.IncomingRequest` (e.g. [packages/next/server/base-server.ts:367](https://github.com/vercel/next.js/commit/3e00a81edeb1da513567456813aa529d92d39d60#diff-2b8e3503c1bebf856610a5551d9c5996435b99f45db389d4eeac1b671518875eL367)). In that commit they started using the `BaseNextRequest` type for the `request` object in the `base-server` file and moved the front-end implementation to `packages/next/server/next-server.ts`, adding a conversion from `http.IncomingMessage` to `BaseNextRequest` with the [`normalizeReq`](https://github.com/vercel/next.js/commit/3e00a81edeb1da513567456813aa529d92d39d60#diff-ea09101c272788873ce720075e4839f0254179af0a07259a107752cafb193949R453-R457) function.

PR #190, opened 2022-05-03
feat: add support for custom request types

wilkmaia:
This PR adds support for a new `option` parameter: `customRequestType`. It represents a custom type from which the `request` object should inherit besides `stream.Readable`.

Closes https://github.com/fastify/light-my-request/issues/189.

#### Checklist

### DIFF
--- a/lib/request.js
+++ b/lib/request.js
@@ -50,6 +50,7 @@ class MockSocket extends EventEmitter {
  * @param {Object} [options.cookies]
  * @param {Object} [options.headers]
  * @param {Object} [options.query]
+ * @param {Object} [options.customRequestType]
  * @param {any} [options.payload]
  */
 function Request (options) {
@@ -119,6 +120,10 @@ function Request (options) {
     payload,
     isDone: false,
     simulate: options.simulate || {}
+  }
+
+  if (options.customRequestType) {
+    util.inherits(this.constructor, options.customRequestType)
   }
 
   return this
--- a/README.md
+++ b/README.md
@@ -162,6 +162,8 @@ Injects a fake request into an HTTP server.
   - `server` - Optional http server. It is used for binding the `dispatchFunc`.
   - `autoStart` - Automatically start the request as soon as the method
     is called. It is only valid when not passing a callback. Defaults to `true`.
+  - `customRequestType` - Optional type from which the `request` object should
+    inherit besides `stream.Readable`
 - `callback` - the callback function using the signature `function (err, res)` where:
   - `err` - error object
   - `res` - a response object where:
--- a/test/test.js
+++ b/test/test.js
@@ -99,6 +99,19 @@ test('request has rawHeaders', (t) => {
   })
 })
 
+test('request has custom type', (t) => {
+  t.plan(2)
+  const dispatch = function (req, res) {
+    t.ok(req instanceof http.IncomingMessage)
+    res.writeHead(200)
+    res.end()
+  }
+
+  inject(dispatch, { method: 'GET', url: 'http://example.com:8080/hello', customRequestType: http.IncomingMessage }, (err, res) => {
+    t.error(err)
+  })
+})
+
 test('passes remote address', (t) => {
   t.plan(2)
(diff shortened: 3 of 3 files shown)


lib/request.js
```diff
@@ -121,6 +122,10 @@ function Request (options) {
     simulate: options.simulate || {}
   }
 
+  if (options.customRequestType) {
+    util.inherits(this.constructor, options.customRequestType)
```
mcollina, line 126: This alters the constructor, which is a singleton property. This means that after the first execution, all the constructors will have this set. Moreover concurrent usage will be broken too.

Could we move this block next to the other util.inherits?

mcollina (changes requested):

lib/request.js
wilkmaia: @mcollina you're right about the constructor. I've updated the code so that custom requests are dealt independently from `Request`.

---

About moving this `util.inherits` call down next to the other one: that'd remove the proposed generality, ie users wouldn't be able to provide specific classes/functions from which the request would inherit. Or am I missing something?

In any case this is more of a curiosity than anything. With the new approach this shouldn't be needed.

lib/request.js
```diff
@@ -50,9 +69,15 @@ class MockSocket extends EventEmitter {
...
  * @param {Object} [options.headers]
  * @param {Object} [options.query]
+ * @param {Object} [options.customRequestType]
  * @param {any} [options.payload]
  */
 function Request (options) {
+  if (options.customRequestType) {
+    assert(options.customRequestType.prototype !== undefined)
+    return new MakeCustomRequest(this, options)
+  }
```
mcollina, line 79: Can you move this if to where Request is instantiated?

mcollina (changes requested):

lib/request.js
wilkmaia: My idea was to keep it transparent to the `lib/request` clients. But sure, will do

mcollina (approved):
lgtm

mcollina:
good work!

climba03003 (commented):
Suggestion only.
Will it be better if the option call `Request` / `RequestPrototype` instead of `customRequestType`?

index.js
```diff
@@ -203,7 +207,11 @@ Object.getOwnPropertyNames(Promise.prototype).forEach(method => {
 })
 
 function isInjection (obj) {
-  return (obj instanceof Request || obj instanceof Response)
+  return (
+    obj instanceof Request ||
+    obj instanceof Response ||
+    obj?.constructor?.name === '_CustomLMRRequest'
```
climba03003, line 213: Optional Chain is not supported by `node@10`
```suggestion
    (obj && obj.constructor && obj.constructor.name === '_CustomLMRRequest')
```

mcollina:
CI is failing

wilkmaia:
> Suggestion only.
> Will it be better if the option call `Request` / `RequestPrototype` instead of `customRequestType`?

I'd be ok with either keeping it as is or updating. I'm usually not great at naming things haha. Instead of `RequestPrototype`, though, I suggest something like `RequestClass` or `RequestType` b/c we expect the base class/constructor, not the prototype itself.

One more thing: did you use pascal case intentionally? Given the other options aren't capitalized wouldn't that be a, yet small, inconsistency in the lib's public api?

climba03003:
> > Suggestion only.
> > Will it be better if the option call `Request` / `RequestPrototype` instead of `customRequestType`?
> 
> I'd be ok with either keeping it as is or updating. I'm usually not great at naming things haha. Instead of `RequestPrototype`, though, I suggest something like `RequestClass` or `RequestType` b/c we expect the base class/constructor, not the prototype itself.
> 
> One more thing: did you use pascal case intentionally? Given the other options aren't capitalized wouldn't that be a, yet small, inconsistency in the lib's public api?

Since it can be a `class` or `function`, I am intended to make it as `PascalCase`. 
If it is `RequestClass`, I would leave it as `Request` only. 
Normally, people would know it should be a `class` if it use `PascalCase`.

Issue #201, opened 2022-07-10
Typing issue with DispatchFunc in TypeScript


segevfiner:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Fastify version

N/A (light-my-request 5.0.0)

### Node.js version

16.15.1

### Operating system

Windows

### Operating system version (i.e. 20.04, 11.3, 10)

10.0.19044.0

### Description

I'm getting the following TypeScript error trying to pass a dispatch func to light-my-request:
```
src/index.ts:12:33 - error TS2345: Argument of type '(req: IncomingMessage, res: ServerResponse) => void' is not assignable to parameter of type 'DispatchFunc'.
  Types of parameters 'req' and 'req' are incompatible.
    Type 'Request' is missing the following properties from type 'IncomingMessage': aborted, httpVersionMajor, httpVersionMinor, complete, and 4 more.

12   const response = await inject(dispatch, {
                                   ~~~~~~~~


Found 1 error in src/index.ts:12
```

A sample repository is linked in Steps to Reproduce.

### Steps to Reproduce

1. Clone https://github.com/segevfiner/light-my-request-typing-issue
2. Run `npm install && npm run build`

### Expected Behavior

I expect to be able to pass dispatch functions that use the plain types from the `http` module.

PR #202, opened 2022-07-11
Add failing test for a plain node http dispatch function

segevfiner:
This is just a failing test for #201, I'm not sure how to fix this, and a fix might be a breaking change in the typings. We seem to have both some variance (Damn I hate variance, so confusing...) issue where TypeScript seems to require `Request` to specify all props of `http.IncomingMessage` for `DispatchFunc` and also mismatches in types between `Request` and `http.IncomingMessage`, such as the `prepare` callback, `url` can be undefined, `methods` is just a `string | undefined` in `http.IncomingMessage` and so on.

#### Checklist

### DIFF
--- a/test/index.test-d.ts
+++ b/test/index.test-d.ts
@@ -90,3 +90,17 @@ const parsedValueUsingGeneric = response.json<ParsedValue>()
 expectType<ParsedValue>(parsedValueUsingGeneric)
 
 expectNotAssignable<http.ServerResponse>(response)
+
+const httpDispatch = function (req: http.IncomingMessage, res: http.ServerResponse) {
+  expectType<boolean>(isInjection(req))
+  expectType<boolean>(isInjection(res))
+
+  const reply = 'Hello World'
+  res.writeHead(200, { 'Content-Type': 'text/plain', 'Content-Length': reply.length })
+  res.end(reply)
+}
+
+inject(httpDispatch, { method: 'get', url: '/' }, (err, res) => {
+  expectType<Error>(err)
+  expectResponse(res)
+})


segevfiner:
We might want to:
* Make `DispatchFunc` just be `http.RequestListener`.
* Make `Request` extend `http.IncomingMessage`, and only include any additional props added by light-my-request.
* Turn `isInjection` into an overloaded [type predicate](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates) to make extra props accessible in the handler with proper types.

That's quite a breaking change though.

mcollina:
Those are all types changes. Everything should just work anyway.

segevfiner:
Yes. For JS users. TS users might have to change their code a bit.

mcollina:
I'm happy to bump a major, in theory none of those type changes are going to impact Fastify: https://github.com/fastify/fastify/blob/83f5dcf2d0aa73e20bdc40da0c37ac7cb2d3b084/types/instance.d.ts#L4

index.d.ts
```diff
@@ -1,4 +1,3 @@
-import * as stream from 'stream'
 import * as http from 'http'
 
 type HTTPMethods = 'DELETE' | 'delete' |
```
segevfiner, line 3: This one actually might cause a bit of a conflict with vanilla typing for many http clients as many just take a `string` for this

index.d.ts
```diff
@@ -60,22 +61,13 @@ declare namespace LightMyRequest {
...
-    headers: http.IncomingHttpHeaders
-    rawHeaders: string[]
-    prepare: (next: () => void) => void
-    // @deprecated
-    connection: object
-  }
+  interface Request extends http.IncomingMessage {}
 
-  interface ServerResponse extends http.ServerResponse {} 
+  interface ServerResponse extends http.ServerResponse {}
```
segevfiner, line 66: I'm not sure what's the real internal name of this object that light-my-request has.

index.d.ts
```diff
@@ -60,22 +61,13 @@ declare namespace LightMyRequest {
...
-    url: string
-    httpVersion: string
-    method: HTTPMethods
-    headers: http.IncomingHttpHeaders
-    rawHeaders: string[]
-    prepare: (next: () => void) => void
-    // @deprecated
-    connection: object
-  }
+  interface Request extends http.IncomingMessage {}
```
segevfiner, line 64: Are there any properties that we consider public that aren't part of the plain `http` objects?

index.d.ts
```diff
@@ -60,22 +61,13 @@ declare namespace LightMyRequest {
...
 
-  interface ServerResponse extends http.ServerResponse {} 
+  interface ServerResponse extends http.ServerResponse {}
 
   interface Response {
     raw: {
-      res: http.ServerResponse,
+      res: ServerResponse,
```
segevfiner, line 70: Is this correct?

index.d.ts
mcollina: This PR did not change this. Am I correct?

index.d.ts
mcollina: I would just drop this interface.

index.d.ts
mcollina: I would just drop this interface.

index.d.ts
segevfiner: Yep. I just wanted to let you know. If you want me to change it, just say so. Otherwise you can always leave changing this for later or not at all.

mcollina: Thanks, that's ok as it is

mcollina (approved):
lgtm

PR #218, opened 2022-08-22
fix: request has aborted flag set to false

### LINKED ISSUE #217: inject hangs when used with Fastify async handler fulfilling undefined
### Prerequisites

### Fastify version

4.x but happening in previous versions as well

### Plugin version

_No response_

### Node.js version

16

### Operating system

Linux

all

### Description

This is an issue related to how this library behaves when used with Fastify. Fastify supports async route handlers and it supports those handlers fulfilling the returned promise with undefined (e.g. not returning anything from the handler).

Example:

```js
fastify.get('/', async () => {
  // implicitly returns a promise fulfilled with undefined
})
```

When `.inject` is used on such routes, it never returns/resolves, and the reason is that Fastify has a special handling for these cases, which is captured ]here](https://github.com/fastify/fastify/blob/d3c3ad78060dd356ef174a45c9585fb36643db7e/lib/wrapThenable.js#L20-L21).

You can see that it handles the behavior of these route handlers so that either they do not resolve undefined, OR the other condition, which requires that:

```js
reply.sent === false && reply.raw.headersSent === false && reply.request.raw.aborted === false
```

The first 2 conditions are satisfied by this library, whereas the third isn't. The fix is fairly straightforward and it's to add that `aborted = false` property to the Request object exposed by this library. Nonetheless before doing this fix I thought it would make sense to discuss whether this change should be done here as it's so specific to how Fastify uses this library.

### Steps to Reproduce

See https://github.com/fastify/fastify/issues/4207

### Expected Behavior

_No response_

simoneb:
Fixes #217. In general though, a more robust approach might be to subclass Node's request object instead of trying to mimic its behavior.

#### Checklist

### DIFF
--- a/lib/request.js
+++ b/lib/request.js
@@ -94,6 +94,7 @@ function Request (options) {
 
   this.url = parsedURL.pathname + parsedURL.search
 
+  this.aborted = false
   this.httpVersionMajor = 1
   this.httpVersionMinor = 1
   this.httpVersion = '1.1'
--- a/test/request.test.js
+++ b/test/request.test.js
@@ -1,0 +1,14 @@
+const { test } = require('tap')
+
+const Request = require('../lib/request')
+
+test('aborted property should be false', async (t) => {
+  const mockReq = {
+    url: 'http://localhost',
+    method: 'GET',
+    headers: {}
+  }
+  const req = new Request(mockReq)
+
+  t.same(req.aborted, false)
+})


mcollina (approved):
lgtm

mcollina:
> In general though, a more robust approach might be to subclass Node's request object instead of trying to mimic its behavior.

I would love to _but_ Node.js APIs does not allow it. `IncomingMessage` is tied to socket networking, and we'd do a lot of monkey patching to get there, which would be even more brittle than this approach.

PR #229, opened 2022-11-24
fix: Response.json() should not check the content-type

### LINKED ISSUE #228: Allow all application/*+json media types to be parsed with response.json()
### 🚀 Feature Proposal

Currently, light-my-request explicitly requires the response's content type to start with exactly "application/json". 

https://github.com/fastify/light-my-request/blob/f895e9a3045da411a30cfe82435b4c284054bb9a/lib/response.js#L142

However, there are several adjacent subtypes, which imply JSON.parse will work, theoretically all:

```
application/[tree "."] subtype +json* [";" parameter]
```

### Motivation

I set the response's content type to "application/geo+json", my `response.json()` calls started throwing errors.

### Example

- https://www.iana.org/assignments/media-types/application/vnd.geo+json
- https://www.iana.org/assignments/media-types/application/geo+json

According to IANA, this applies to roughly 100 more media types: [Source](https://www.iana.org/assignments/media-types/media-types.xhtml)

WoH:
Closes #228 

#### Checklist

### DIFF
--- a/lib/response.js
+++ b/lib/response.js
@@ -139,8 +139,8 @@ function generatePayload (response) {
 
   // Prepare payload parsers
   res.json = function parseJsonPayload () {
-    if (res.headers['content-type'].indexOf('application/json') < 0) {
-      throw new Error('The content-type of the response is not application/json')
+    if (!/^application\/(?<tree>([\w-]*\.)*)(?<subtype>[\w-]*\+)?json(.*)$/.test(res.headers['content-type'])) {
+      throw new Error('The content-type of the response is not application/json or a related media type')
     }
     return JSON.parse(res.payload)
   }
--- a/lib/response.js
+++ b/lib/response.js
@@ -139,9 +139,6 @@ function generatePayload (response) {
 
   // Prepare payload parsers
   res.json = function parseJsonPayload () {
-    if (!/^application\/(?<tree>([\w-]*\.)*)(?<subtype>[\w-]*\+)?json(.*)$/.test(res.headers['content-type'])) {
-      throw new Error('The content-type of the response is not application/json or a related media type')
-    }
     return JSON.parse(res.payload)
   }
 
--- a/README.md
+++ b/README.md
@@ -181,7 +181,7 @@ Injects a fake request into an HTTP server.
     - `body` - alias for payload.
     - `rawPayload` - the raw payload as a Buffer.
     - `trailers` - an object containing the response trailers.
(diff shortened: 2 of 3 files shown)


mcollina (approved):
lgtm

lib/response.js
```diff
@@ -139,8 +139,8 @@ function generatePayload (response) {
 
   // Prepare payload parsers
   res.json = function parseJsonPayload () {
-    if (res.headers['content-type'].indexOf('application/json') < 0) {
-      throw new Error('The content-type of the response is not application/json')
+    if (!/^application\/(?<tree>([\w-]*\.)*)(?<subtype>[\w-]*\+)?json(.*)$/.test(res.headers['content-type'])) {
```
climba03003, line 142: I am thinking, would it be better to remove the check entirely.

The function here is more or less the same as `fetch`. And it should not block the parsing based on `Content-Type`.

WoH: I am happy to remove it entirely in case that's a better way to approach it. But I assume this is not supposed to get too close to fetch (i.e. adding the content type if it wasn't set)?

climba03003: I means `.json` is just doing `JSON.parse` which is a helper function only (same as `fetch`).

climba03003: Let see what the others think about it. 
@fastify/core 

I am also fine for the current implementation. @WoH And it is missing document update.
https://github.com/fastify/light-my-request/blob/f895e9a3045da411a30cfe82435b4c284054bb9a/README.md?plain=1#L184

WoH: Thanks for the hint, let me know if my attempts at an explanation that doesn't sound too confusing worked 😅

Uzlopak: Looks kind of a potential performance bottleneck. Also why are there named groups if you just use .test() and no .match()? 

Also safe-regex shows that the regex is potentially unsafe.

WoH: Groups to have a better idea what is going on. The current benchmarks ignore this, so posting them didn't seem helpful, I assume 'application/json' is the most common content type by far, maybe check that first and skip the regex?

mcollina: I would just drop the check to be honest.

kibertoad: +1 to dropping the check

climba03003 (approved):

lib/response.js
WoH: Dropped the check, let me know it if I should change anything else.

RafaelGSS (approved):

mcollina (approved):
lgtm

kibertoad (approved):

PR #234, opened 2022-12-05
add response cookies type

trim21:
#### Checklist

type only change, benchmark is not necessary

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -58,6 +58,20 @@ declare namespace inject {
     Request?: object,
   }
 
+  /**
+   * https://github.com/nfriedly/set-cookie-parser/blob/3eab8b7d5d12c8ed87832532861c1a35520cf5b3/lib/set-cookie.js#L41
+   */
+  interface Cookie {
+    name: string;
+    value: string;
+    expires?: Date;
+    maxAge?: number;
+    secure?: boolean;
+    httpOnly?: boolean;
+    sameSite?: string;
+    [name: string]: unknown
+  }
+
   export interface Response {
     raw: {
       res: http.ServerResponse,
@@ -71,7 +85,7 @@ declare namespace inject {
     payload: string
     body: string
     json: <T = any>() => T
-    cookies: Array<object>
+    cookies: Array<Cookie>
   }
 
   export interface Chain extends Promise<Response> {
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -69,7 +69,7 @@ declare namespace inject {
     secure?: boolean;
     httpOnly?: boolean;
     sameSite?: string;
-    [key: string]: string;
+    [name: string]: unknown
   }
 
   export interface Response {
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -69,7 +69,7 @@ declare namespace inject {
     secure?: boolean;
     httpOnly?: boolean;
     sameSite?: string;
-    [name: string]: unknown
+    [key: string]: string;
   }
 
   export interface Response {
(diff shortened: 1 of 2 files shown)


trim21:
This PR change only type of `Response.cookies`

I'm writing test code using cookies, it's like this:

```ts
    const res = await app.inject({
      url: '/v1/login',
      method: 'post',
      payload: { ... },
    });

    expect(res.statusCode).toBe(200);
    expect(res.json()).toEqual(treeHoleUser);

    console.log(res.cookies);

    const cookieValue = res.cookies.find((x) => x.name === 'sessionID')?.value; // type error here
```

And it cause a little type problem when trying to get cookies value, so this PR add type to `Response.cookies`

types/index.d.ts
```diff
@@ -58,6 +58,20 @@ declare namespace inject {
...
+   */
+  interface Cookie {
+    name: string;
+    value: string;
+    expires?: Date;
+    maxAge?: number;
+    secure?: boolean;
+    httpOnly?: boolean;
+    sameSite?: string;
+    [name: string]: unknown
```
Uzlopak, line 72:
```suggestion
    [key: string]: string
```

It is not unknown, but always string ;)

types/index.test-d.ts
```diff
@@ -21,6 +21,7 @@ const expectResponse = function (res: Response) {
   expectAssignable<Function>(res.json)
   expectAssignable<http.ServerResponse>(res.raw.res)
   expectAssignable<http.IncomingMessage>(res.raw.req)
+  expectAssignable<Record<string, string>>(Object.fromEntries(res.cookies.map(x => [x.name, x.value])))
```
Uzlopak, line 24: very complicated...  and only checks name and value

why not 

expectType<Date | undefined>(res.cookies[0].expires

types/index.d.ts
trim21: Yes not it can't be string here

trim21: `string` conflict with known non-string property like `httpOnly`

trim21: so it will be `unknown` or `any` or `string|boolean|Date`

trim21: when you get property with index like `cookies['httpOnly']`, ts will return type `unknown` instead known `boolean`

mcollina (approved):
lgtm

types/index.test-d.ts
```diff
@@ -21,6 +21,7 @@ const expectResponse = function (res: Response) {
   expectAssignable<Function>(res.json)
   expectAssignable<http.ServerResponse>(res.raw.res)
   expectAssignable<http.IncomingMessage>(res.raw.req)
+  expectType<Array<Date | undefined>>(res.cookies.map(x => x.expires))
```
Uzlopak, line 24: As this is a PR to improve the typings,we should test intensively the typings:

```suggestion
  expectAssignable<Array<any>>(res.cookies)
  expectType<string>(res.cookies[0].name)
  expectType<string>(res.cookies[0].value)
  expectType<Date | undefined>(res.cookies[0].expires)
  expectType<number | undefined>(res.cookies[0].maxAge)
  expectType<boolean | undefined>(res.cookies[0].httpOnly)
  expectType<boolean | undefined>(res.cookies[0].secure)
  expectType<string | undefined>(res.cookies[0].sameSite)
  expectType<unknown | undefined>(res.cookies[0].additional)
```

trim21: I assumed tsd would enable `noUncheckedIndexedAccess` so I want to avoid using `.cookies[0]`, but now looks like it doesn't ...

Uzlopak (approved):
thanks

trim21:
Is there a release cycle? 🤔 When can I expect this change being released on npm after it's merged

PR #244, opened 2023-05-04
test: add multipart test

Eomm:
ref: https://github.com/fastify/help/issues/867


mcollina (changes requested):
This creates something of a dependency loop. Maybe just use busboy?

PR #250, opened 2023-06-24
dont serialize null as header values

### LINKED ISSUE #249: Headers with value set to null are wrongly emitted as string in the response
### Prerequisites

### Fastify version

4.18.0

### Plugin version

5.10.0

### Node.js version

16.x and above

### Operating system

Linux

20.04

### Description

Our test suite, based on the test application provided by fastify, is suddenly failing even though we did not change fastify version.

This is likely because of this change:

https://github.com/fastify/light-my-request/commit/23f4fa06d0748448e045bfb28f500959f1550a49

Before, setting a response header value to `null` in the route handler was successfully propagated as `null` in the response returned by fastify test application.

Now, setting a response header value to `null` in the route handler leads to the string "null" being propagated in the response returned by fastify test application. Making our test suite fails.

This is a breaking change: this library has always been supporting setting a header to `null`. It does not support it anymore.

### Steps to Reproduce

Just set a header value to `null` using the `header` or `headers` function of fastify response and check the value of the header in the response returned by the test application. It will be the string "null" instead of the `null` value.

### Expected Behavior

_No response_

Uzlopak:
fixes #249 

#### Checklist

### DIFF
--- a/lib/response.js
+++ b/lib/response.js
@@ -162,6 +162,8 @@ function serializeHeaders (response) {
     const headerValue = headers[headerName]
     if (Array.isArray(headerValue)) {
       headers[headerName] = headerValue.map(value => '' + value)
+    } else if (headerValue === null || headerValue === undefined) {
+      delete headers[headerName]
     } else {
       headers[headerName] = '' + headerValue
     }
--- a/test/index.test.js
+++ b/test/index.test.js
@@ -1620,7 +1620,6 @@ test('correctly handles no string headers', (t) => {
     t.strictSame(res.headers, {
       integer: '12',
       float: '3.14',
-      null: 'null',
       string: 'string',
       object: '[object Object]',
       array: ['1', 'two', '3'],


Fdawgs (approved):
This got me as I use `undefined` in my header tests, thanks for the fix!

ivan-tymoshenko (approved):

mcollina:
Technically this does not match node core behavior. It serializes `null` as `'null'`, and this module mimics core as much as possible.

Fdawgs:
> Technically this does not match node core behavior. It serializes `null` as `'null'`, and this module mimics core as much as possible.

Ah balls, what about `undefined`?

Fdawgs:
> > Technically this does not match node core behavior. It serializes `null` as `'null'`, and this module mimics core as much as possible.
> 
> Ah balls, what about `undefined`?

@mcollina @Uzlopak thoughts on this?

mcollina:
```js
const { createServer, request } = require('http');

const server = createServer((req, res) => {
  res.setHeader('Content-Type', 'text/plain');

  try {
    res.setHeader('X-undefined', undefined);
  } catch (err) {
    console.log('undefined could not be set')
  }
  res.setHeader('X-null', null);
  res.end('Hello World!');
})

server.listen(3000, function () {
  console.log('Server listening on port 3000');
  request('http://localhost:3000', function (res) {
    console.log(res.headers);
    process.exit(0);
  }).end()
});
```

Fdawgs:
Aight, cool.

I was using `undefined` in my tests like so to remove expected headers that I... no longer expected:
``` js 
const sitewideExpectedHeaders = {
	  'Content-Type': 'application/json',
	  expires: "0",
}

const cacheablepageExpectedHeaders = {
	...expectedHeaders,
	expires: undefined
}
```

Guess I'll need to just change it to use `delete cacheablepageExpectedHeaders.expires` instead.

PR #252, opened 2023-08-31
migrate to class

sirenkovladd:
migrate to class syntax (Chain, request, response)

bun.js support
update usage of deprecated attribute
- [http.ServerResponse._header](https://github.com/fastify/light-my-request/blob/v5.10.0/lib/response.js#L177)
- [http.ServerResponse.connection](https://github.com/fastify/light-my-request/blob/v5.10.0/lib/response.js#L45)

#### Checklist

- [X] run `npm run test` and `npm run benchmark`
- [X] tests and/or benchmarks are included
- [X] documentation is changed or added (.d.ts)
- [X] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/master/CONTRIBUTING.md#developers-certificate-of-origin-11)
      and the [Code of conduct](https://github.com/fastify/.github/blob/master/CODE_OF_CONDUCT.md)

PR #255, opened 2023-09-03
Parsing socket

sirenkovladd:
migrate to class syntax (Chain, request, response)
add a mock socket class (parse headers, body, trailers)
change emitting and subscription

this is the sequel to https://github.com/fastify/light-my-request/pull/252

bun.js support
update usage of deprecated attribute
- [http.ServerResponse._header](https://github.com/fastify/light-my-request/blob/v5.10.0/lib/response.js#L177)
- [http.ServerResponse.connection](https://github.com/fastify/light-my-request/blob/v5.10.0/lib/response.js#L45)

some test was changed due to unexpected behavior
the main tests were tested with
```js
http.createServer(inj).listen(3000, () => {
  const r = http.request({ port: 3000, method: 'POST' }, (res) => cb(null, res)).on('error', cb)
  payload && payload.pipe(r)
  r.end()
});
``` 

#### Checklist

- [X] documentation is changed or added (.d.ts)
- [X] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/master/CONTRIBUTING.md#developers-certificate-of-origin-11)

### DIFF
--- a/lib/chain.js
+++ b/lib/chain.js
@@ -1,0 +1,107 @@
+'use strict'
+
+const doInject = require('./do-inject')
+
+const errorMessage = 'The dispatch function has already been invoked'
+
+class Chain {
+  _hasInvoked = false
+  _promise = null
+  option
+  dispatch
+
+  constructor (dispatch, option) {
+    this.dispatch = dispatch
+    if (typeof option === 'string') {
+      this.option = { url: option }
+    } else {
+      this.option = Object.assign({}, option)
+    }
+
+    if (this.option.autoStart !== false) {
+      process.nextTick(() => {
+        if (!this._hasInvoked) {
+          this.end()
+        }
+      })
+    }
+  }
+
+  /**
+   * @private
+   * @param {string} method
+   * @param {string} url
+   */
+  wrapHttpMethod (method, url) {
+    if (this._hasInvoked === true || this._promise) {
+      throw new Error(errorMessage)
+    }
+    this.option.url = url
+    this.option.method = method.toUpperCase()
+    return this
+  }
+
+  delete (url) { return this.wrapHttpMethod('delete', url) }
+  get (url) { return this.wrapHttpMethod('get', url) }
+  head (url) { return this.wrapHttpMethod('head', url) }
+  options (url) { return this.wrapHttpMethod('options', url) }
+  patch (url) { return this.wrapHttpMethod('patch', url) }
+  post (url) { return this.wrapHttpMethod('post', url) }
+  put (url) { return this.wrapHttpMethod('put', url) }
+  trace (url) { return this.wrapHttpMethod('trace', url) }
+
+  /**
+   * @private
+   * @param {string} method
+   * @param {string} url
+   */
+  wrapChainMethod (method, value) {
+    if (this._hasInvoked === true || this._promise) {
+      throw new Error(errorMessage)
+    }
+    this.option[method] = value
+    return this
+  }
+
+  body (url) { return this.wrapChainMethod('body', url) }
+  cookies (url) { return this.wrapChainMethod('cookies', url) }
+  headers (url) { return this.wrapChainMethod('headers', url) }
+  payload (url) { return this.wrapChainMethod('payload', url) }
+  query (url) { return this.wrapChainMethod('query', url) }
+
+  end (callback) {
+    if (this._hasInvoked === true || this._promise) {
+      throw new Error(errorMessage)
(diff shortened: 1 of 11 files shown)


sirenkovladd:
It takes some time to optimize
but generally open to discussion

sirenkovladd:
```sh
light-my-request on  parsing-socket [!+⇕] is 📦 v5.10.0 via  v20.5.0 on ☁️  (us-west-2) took 5s 
❯ npm run benchmark

> light-my-request@5.10.0 benchmark
> node benchmark/benchmark.js

Request x 160,977 ops/sec ±3.86% (80 runs sampled)
Custom Request x 14,371 ops/sec ±4.12% (81 runs sampled)
Request With Cookies x 136,650 ops/sec ±1.94% (87 runs sampled)
Request With Cookies n payload x 134,341 ops/sec ±2.71% (81 runs sampled)
ParseUrl x 1,207,543 ops/sec ±1.39% (87 runs sampled)
ParseUrl and query x 114,550 ops/sec ±2.00% (90 runs sampled)
```

mcollina (changes requested):
There seems to be quite a significant performance regression with this PR.

I _think_ it's due to the addition of the internal Socket, so I think this might be a no-go.

I wouldn't also claim bun support without the tests running on Bun too.

types/index.d.ts
```diff
@@ -14,7 +14,7 @@ declare namespace inject {
 
   export type DispatchFunc = http.RequestListener
 
-  export type CallbackFunc = (err: Error, response: Response) => void
+  export type CallbackFunc = (err: Error | undefined, response: Response | undefined) => void
```
Eomm, line 17: This could be a dedicated PR

test/index.test.js
```diff
@@ -723,16 +750,16 @@ test('can override stream payload content-length header', (t) => {
   const headers = { 'content-length': '100' }
 
   inject(dispatch, { method: 'POST', url: '/', payload: getTestStream(), headers }, (err, res) => {
-    t.error(err)
-    t.equal(res.payload, '100')
+    t.equal(err instanceof inject.errors.ContentLength, true)
+    t.error(res)
```
Eomm, line 754: The test says `can override stream payload content-length header` - are we blocking this feature now?

test/index.test.js
```diff
@@ -806,14 +831,14 @@ test('simulates split', (t) => {
   })
 })
 
-test('simulates error', (t) => {
+t.skip('simulates error', (t) => {
```
Eomm, line 834:
```suggestion
test('simulates error', (t) => {
```

lib/chain.js
```diff
@@ -0,0 +1,107 @@
+'use strict'
+
+const doInject = require('./do-inject')
+
+const errorMessage = 'The dispatch function has already been invoked'
+
+class Chain {
```
Eomm, line 7: This refactor could be another PR - smaller PRs helps to get the code merged faster

Eomm (commented):
It is hard to find the old code and the new code feature.

I looked at the commits to follow the step-by-step process, but 2 commits did not help.

I really suggest splitting this PR into smaller pieces

mcollina:
I don't think we agreed to https://github.com/fastify/light-my-request/pull/252, this PR seems a bit premature too.

types/index.d.ts
sirenkovladd: https://github.com/fastify/light-my-request/pull/256

test/index.test.js
sirenkovladd: Yes
1. it is not supported with test example http.Server/http.request
2. I can't call the Dispatch function unless I read everything from Readable

test/index.test.js
sirenkovladd: This was done on purpose as it could not be reproduced with a real server and request
maybe you have ideas how to do it?

sirenkovladd:
> I don't think we agreed to https://github.com/fastify/light-my-request/pull/252, this PR seems a bit premature too.

Yes, so I made a another PR with planning for the long future

sirenkovladd:
This PR will simply allow it to be more native, with support for most events and most attribute, methods
that are currently available in native request Node but not in light-my-request

that's why I used extends Socket and IncomingMessage

Tony133:
Closing this PR as it has been open for quite some time and is now heavily conflicting with the current codebase.

If this is still relevant, please open a new PR based on the latest codebase.

PR #317, opened 2024-12-03
Streaming input

mcollina:
This PR does two things:

1. if a stream is provided as the body of the request, it does not accumulate it anymore but forward it respecting backpressure instead.
2. fix an old incosistency with non-fake servers that had a response being sent even if the dispatched handler aborted

#### Checklist

- [x] passes all Fastify tests


index.js
```diff
@@ -16,16 +16,40 @@ function inject (dispatchFunc, options, callback) {
...
+    return next()
+  }
+
+  const chunks = []
+
+  payload.on('data', (chunk) => chunks.push(Buffer.from(chunk)))
+
+  payload.on('end', () => {
+    const payload = Buffer.concat(chunks)
+    req.headers['content-length'] = req.headers['content-length'] || ('' + payload.length)
```
ShogunPanda, line 32: Make sure you delete the eventual `transfer-encoding` header here since you're setting the content length.

lib/request.js
```diff
@@ -193,34 +193,46 @@ function Request (options) {
     addAbortSignal(signal, this)
   }
 
-  return this
-}
+  {
+    const payload = this._lightMyRequest.payload
+    if (payload && payload._readableState) { // does not quack like a stream
```
ShogunPanda, line 198: I think you mean "quacks like a stream" here :)

ShogunPanda (approved):
Small nits, but LGTM!

mcollina:
@fastify/libraries I need another review

lib/request.js
jsumners: I would just fix this comment.

jsumners (approved):

PR #333, opened 2025-01-28
fix: restore express prototypes so injecting doesn't break http requests

gunjam:
I'm attempting to port an internal express app test framework to use light-my-request because it hugely increases performance (thanks!), but some of the framework's plugins still need to make HTTP requests.

Currenlty the first `inject()` call will permenantly modify the app's request and response prototypes, making subsequent HTTP requests crash the server - this change fixes that by caching them and setting the back to the originals.

Ta

#### Checklist

### DIFF
--- a/index.js
+++ b/index.js
@@ -74,8 +74,21 @@ function doInject (dispatchFunc, options, callback) {
 
   // Express.js detection
   if (dispatchFunc.request && dispatchFunc.request.app === dispatchFunc) {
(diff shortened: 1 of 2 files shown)


mcollina (approved):
lgtm

mcollina:
I had to revert this due to a test breaking. https://github.com/fastify/light-my-request/actions/runs/13565121018/job/37916398433

Can you take another look?

PR #339, opened 2025-02-26
Add `req.socket.destroy()`

yus-ham:
#### Checklist

- [x] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/main/CONTRIBUTING.md#developers-certificate-of-origin-11)

This ensures invocation to `req.socket.destroy()` won't makes exception thrown as it happen before in my express app
```sh
125 |     // cannot actually respond
126 |     if (headersSent(res)) {
127 |       debug('cannot %d after headers sent', status)
128 |       if (req.socket) {
129 |         console.info('finalhandler req.socket:',req.socket)
130 |         req.socket.destroy()
                         ^
TypeError: req.socket.destroy is not a function. (In 'req.socket.destroy()', 'req.socket.destroy' is undefined)
      at <anonymous> (/app/node_modules/finalhandler/index.js:130:20)
```

### DIFF
--- a/lib/request.js
+++ b/lib/request.js
@@ -42,6 +42,9 @@ class MockSocket extends EventEmitter {
     super()
     this.remoteAddress = remoteAddress
   }
+
+  destroy () {
+  }
 }
 
 /**
--- a/test/request.test.js
+++ b/test/request.test.js
@@ -14,3 +14,10 @@ test('aborted property should be false', async (t) => {
 
   t.assert.strictEqual(req.aborted, false)
 })
+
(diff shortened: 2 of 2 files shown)


mcollina (approved):
lgtm

jsumners (commented):
This seems reasonable, but I wonder what the use case is that makes it necessary. This library has been around for a long time without any reports of this being a problem.

Basically, a test has been included to assert that the method exists and doesn't do anything, but there is not a test that shows what is being fixed.

PR #388, opened 2026-07-14
fix: aggregate array request header values like a real HTTP server

### LINKED ISSUE #264: Multi-value request headers are serialized to comma-separated string
### Prerequisites

### Fastify version

4.23.2

### Plugin version

_No response_

### Node.js version

20.x

### Operating system

macOS

13.0

### Description

The `.inject()` method allows to inject request headers which values can be a string or an array of strings. When an array is used, it gets serialized to an comma-separated string.

### Steps to Reproduce

 ```
 await fastifyInstance.inject({ method: 'get', url: '/', headers: { foo: 'bar', baz: ['first', 'second'] } });
```

gives me the following request headers in the route handler:

```
{
  baz: 'first,second',
   foo: 'bar',
  host: 'localhost:80',
   'user-agent': 'lightMyRequest'
}
```

### Expected Behavior

I would expect to get an array of strings in the route handler as well.

### PR DESCRIPTION
**Pull request by LeSingh1**
## Problem

`inject()` accepts array-valued request headers (allowed by the `http.OutgoingHttpHeaders` input type) to simulate the same header being sent multiple times. However, array values were serialized with a bare `'' + value` coercion, which does not match how a real Node.js HTTP server aggregates repeated header lines onto `request.headers` / `request.rawHeaders`.

Given:

```js
inject(dispatch, {
  method: 'GET',
  url: '/',
  headers: { baz: ['first', 'second'], 'set-cookie': ['a=1', 'b=2'], cookie: ['x=1', 'y=2'] }
})
```

| | light-my-request (before) | Real Node.js HTTP server |
| --- | --- | --- |
| `headers.baz` | `"first,second"` | `"first, second"` |
| `headers['set-cookie']` | `"a=1,b=2"` | `["a=1", "b=2"]` |
| `headers.cookie` | `"x=1,y=2"` | `"x=1; y=2"` |
| `rawHeaders` | `["baz","first,second", ...]` | `["baz","first","baz","second", ...]` |

The Node behavior in the right column was verified empirically against `http.createServer` receiving duplicate header lines over a raw socket. See issue #264.

## Fix

Aggregate array header values exactly as `http.IncomingMessage` does (per the [`message.headers`](https://nodejs.org/api/http.html#messageheaders) docs):

- `set-cookie` is kept as an array of its individual values.
- `cookie` values are joined with `'; '`.
- Duplicates of single-value headers (`content-type`, `host`, `authorization`, …) are discarded, keeping the first.
- All other headers are joined with `', '`.
- Every supplied value is preserved as a separate `rawHeaders` entry.

Non-array (string/number) header values are unchanged, so existing behavior is fully preserved for the common case. One existing test that asserted the old bare-comma output (`'1,two,3'`) was updated to the Node-faithful `'1, two, 3'`.

## Verification

- Added regression tests covering the `', '` join + `rawHeaders` expansion, `set-cookie` array preservation, `cookie` `'; '` join, single-value-header duplicate discarding, and the `undefined`-inside-array assertion.
- `npm test` passes: lint, **172** unit tests at **100%** coverage (c8 `--100`), and **54** tstyche type assertions.

Closes #264

### DIFF
--- a/README.md
+++ b/README.md
@@ -152,7 +152,7 @@ Injects a fake request into an HTTP server.
   - `method` - a string specifying the HTTP request method, defaulting to `'GET'`.
   - `authority` - a string specifying the HTTP HOST header value to be used if no header is provided, and the `url`
     does not include an authority component. Defaults to `'localhost'`.
-  - `headers` - an optional object containing request headers.
+  - `headers` - an optional object containing request headers. A header value may be an array of strings to simulate the same header being sent multiple times; the values are aggregated exactly as a Node.js HTTP server would expose them on `request.headers` (`set-cookie` is kept as an array, `cookie` values are joined with `'; '`, other values with `', '`), while every individual value is preserved in `request.rawHeaders`.
   - `cookies` - an optional object containing key-value pairs that will be encoded and added to `cookie` header. If the header is already set, the data will be appended.
   - `remoteAddress` - an optional string specifying the client remote address. Defaults to `'127.0.0.1'`.
   - `payload` - an optional request payload. Can be a string, Buffer, Stream, or object. If the payload is string, Buffer or Stream is used as is as the request payload. Otherwise, it is serialized with `JSON.stringify` forcing the request to have the `Content-type` equal to `application/json`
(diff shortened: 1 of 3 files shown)


mcollina (approved):
lgtm

Tony133 (approved):
lgtm

PR #391, opened 2026-07-18
fix(types): allow custom HTTP methods in InjectOptions.method

### LINKED ISSUE #385: Support custom HTTP method (TypeScript)
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have searched existing issues to ensure the feature has not already been requested

### 🚀 Feature Proposal

I fail to test custom HTTP method added to my server since `InjectOptions.method` type is a closed literal union.

After adding a custom HTTP method using fastify server `addHttpMethod` [method](https://fastify.dev/docs/v5.3.x/Guides/Migration-Guide-V5/?_highlight=ddhttpmethod#removal-of-some-non-standard-http-methods), using `server.inject` with this method throw a type-check error.

### Example

```
fastify.addHttpMethod('REBIND', { hasBody: true })
...

await fastify.inject({ method: 'REBIND' ... }); //  Type '"REBIND"' is not assignable to type 'HTTPMethods'
```

tzafrirben:
Widen `HTTPMethods` type to accept any string via `(string & {})` alongside the known literals, so methods registered with Fastify's addHttpMethod (e.g. 'QUERY', 'REBIND') type-check while autocomplete
for standard methods is preserved.

Fixes #385

### The change
Use the standard TypeScript "literal union + string" to widen `HTTPMethods` type in types/index.d.ts 

The (string & {}) intersection prevents TypeScript from collapsing the union down to plain string (which would lose IDE autocomplete for the known method literals), while still allowing any other string value — like a custom method registered via Fastify's addHttpMethod — to type-check.

#### Checklist

- [ ] run `npm run test && npm run benchmark --if-present`
- [ ] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/main/CONTRIBUTING.md#developers-certificate-of-origin-11)

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -7,7 +7,8 @@ type HTTPMethods = 'DELETE' | 'delete' |
                    'PATCH' | 'patch' |
                    'POST' | 'post' |
                    'PUT' | 'put' |
-                   'OPTIONS' | 'options'
+                   'OPTIONS' | 'options' |
+                   (string & {})
 
 type Inject = typeof inject
 
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -8,6 +8,7 @@ type HTTPMethods = 'DELETE' | 'delete' |
                    'POST' | 'post' |
                    'PUT' | 'put' |
                    'OPTIONS' | 'options' |
+                   'QUERY' | 'query' |
                    (string & {})
 
 type Inject = typeof inject
--- a/types/index.tst.ts
+++ b/types/index.tst.ts
@@ -10,7 +10,7 @@ expect({ autoStart: true }).type.toBeAssignableTo<InjectOptions>()
 expect({ autoStart: false }).type.toBeAssignableTo<InjectOptions>()
 expect({ validate: true }).type.toBeAssignableTo<InjectOptions>()
 expect({ validate: false }).type.toBeAssignableTo<InjectOptions>()
-expect({ method: 'QUERY', url: '/' }).type.toBeAssignableTo<InjectOptions>()
+expect({ method: 'REBIND', url: '/' }).type.toBeAssignableTo<InjectOptions>()
 expect({ method: 'REBIND' }).type.toBeAssignableTo<InjectOptions>()
 
 const dispatch: http.RequestListener = function (req, res) {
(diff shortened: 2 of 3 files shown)


types/index.d.ts
```diff
@@ -7,7 +7,8 @@ type HTTPMethods = 'DELETE' | 'delete' |
                    'PATCH' | 'patch' |
                    'POST' | 'post' |
                    'PUT' | 'put' |
-                   'OPTIONS' | 'options'
+                   'OPTIONS' | 'options' |
+                   (string & {})
```
jean-michelet, line 12: Well, if we allow any custom http method, let set `HTTPMethods: string`

types/index.tst.ts
```diff
@@ -10,6 +10,7 @@ expect({ autoStart: true }).type.toBeAssignableTo<InjectOptions>()
 expect({ autoStart: false }).type.toBeAssignableTo<InjectOptions>()
 expect({ validate: true }).type.toBeAssignableTo<InjectOptions>()
 expect({ validate: false }).type.toBeAssignableTo<InjectOptions>()
+expect({ method: 'QUERY', url: '/' }).type.toBeAssignableTo<InjectOptions>()
```
jean-michelet: Don't know if it exists yet in the test suite, take a look plz, but I would ensure there is a runtime test proving any custom method can be used.

types/index.d.ts
tzafrirben: I though about it - however, keeping the literal union will allow IDE autocomplete and suggest user with standard verb. The autocomplete will be lost if we collapse it to `string` .

Saying that, if there's a downside to keep the union I can change it to string

jean-michelet: Ok, so I think we can add `QUERY` as a default then: https://github.com/fastify/fastify/pull/6832#issue-4810495252

types/index.tst.ts
tzafrirben: I have added a runtime test using `REBIND` method to prove custom HTTP methods work end-to-end with default validation

types/index.d.ts
tzafrirben: `QUERY/query` was added explicitly to `HTTPMethods`, matching fastify's approach in [fastify/fastify#6832](https://github.com/fastify/fastify/pull/6832#issue-4810495252)

types/index.tst.ts
```diff
@@ -10,6 +10,8 @@ expect({ autoStart: true }).type.toBeAssignableTo<InjectOptions>()
...
```
jean-michelet:
```suggestion
expect({ method: 'REBIND', url: '/' }).type.toBeAssignableTo<InjectOptions>()
```

jean-michelet: And it will be good for me

jean-michelet (approved):

mcollina (approved):
lgtm

PR #401, opened 2026-08-22
feat: implements injected cookies encoding respecting http state management rfc

big-kahuna-burger:
## Summary
- `inject({ cookies })` now encodes cookies onto the `Cookie` header using [RFC 6265 §5.1.4](https://datatracker.ietf.org/doc/html/rfc6265#section-5.1.4) path matching, so a cookie with `Path=/account/123` is not sent to `/account/other`.
- Accepts the existing `name: value` map (no path → treated as `Path=/`), `name: { value, path }` records, or a previous `res.cookies` array.
- Docs live in `docs/cookies.md` (so I don't pollute a nice readme with extra hunk related to cookies only); the generated `lib/config-validator.js` allows a cookie array as well as an object.

## Test plan
- [x] `npm test` (lint, 100% unit coverage, tstyche)
- [x] Confirm string cookies still go out on every path: `{ session: 'abc' }`
- [x] Confirm a scoped cookie is omitted when the request path does not match
- [x] Confirm a scoped cookie is sent when the request path is a valid RFC 6265 prefix (`/account/123` → `/account/123/settings`, not `/account/other` or `/docsets` vs `/docs`)
- [x] Replay `login.cookies` onto a later `inject` and check only matching cookies appear on `Cookie`

#### Checklist

- [x] run `npm run test && npm run benchmark --if-present`
- [x] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/main/CONTRIBUTING.md#developers-certificate-of-origin-11)

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,7 +153,7 @@ Injects a fake request into an HTTP server.
   - `authority` - a string specifying the HTTP HOST header value to be used if no header is provided, and the `url`
     does not include an authority component. Defaults to `'localhost'`.
   - `headers` - an optional object containing request headers. A header value may be an array of strings to simulate the same header being sent multiple times; the values are aggregated exactly as a Node.js HTTP server would expose them on `request.headers` (`set-cookie` is kept as an array, `cookie` values are joined with `'; '`, other values with `', '`), while every individual value is preserved in `request.rawHeaders`.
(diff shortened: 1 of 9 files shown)


mcollina (approved):
lgtm

Tony133 (approved):

